PR #3, opened 2024-07-01
lib: Add transport and hsm fixes

slingappa:


lib/rpmi_hsm.c
```diff
@@ -207,7 +207,7 @@ static void __rpmi_hsm_process_hart_state_changes(struct rpmi_hsm *hsm,
 		return;
 
 	hw_state = hsm->leaf.ops->hart_get_hw_state(hsm->leaf.ops_priv, hart_index);
-	if (hart->state < 0) {
+	if ((rpmi_uint32_t)hart->state < 0) {
```
avpatel, line 210: Why this change is required ?

lib/rpmi_transport_shmem.c
```diff
@@ -146,7 +146,7 @@ static enum rpmi_error shmem_dequeue(struct rpmi_transport *trans,
 	}
 
 	headidx = rpmi_to_le32(rpmi_env_mod32(headidx + 1, shtrans->data_slots));
-	rc = rpmi_shmem_write(shmem, queue_base + trans->slot_size,
+	rc = rpmi_shmem_write(shmem, queue_base,
```
avpatel, line 149: Good catch.

lib/rpmi_transport_shmem.c
```diff
@@ -204,7 +204,7 @@ struct rpmi_transport *rpmi_transport_shmem_create(const char *name,
 		return NULL;
 
 	shtrans->shmem = shmem;
-	shtrans->queue_size = rpmi_shmem_size(shmem) / RPMI_QUEUE_MAX;
+	shtrans->queue_size = rpmi_shmem_size(shmem) / RPMI_QUEUE_MAX / 2;
```
avpatel, line 207: This change is not correct.

The shared memory provided to `rpmi_transport_shmem_create()` will be consumed entirely by the 4 queues. If some part of the shared memory is to be used as fast channel then the caller should create `rpmi_shmem` instance accordingly.

lib/rpmi_service_group_hsm.c
```diff
@@ -170,8 +170,8 @@ static enum rpmi_error rpmi_hsm_sg_get_hart_list(struct rpmi_service_group *grou
 
 	*response_datalen = (returned + 3) * sizeof(*resp);
 	resp[0] = rpmi_to_xe32(trans->is_be, (rpmi_uint32_t)status);
-	resp[1] = rpmi_to_xe32(trans->is_be, returned);
-	resp[2] = rpmi_to_xe32(trans->is_be, remaining);
+	resp[1] = rpmi_to_xe32(trans->is_be, remaining);
+	resp[2] = rpmi_to_xe32(trans->is_be, returned);
```
avpatel, line 174: Good catch.

lib/rpmi_hsm.c
slingappa: hart->state is enum, that being 0xFFFFFFFF(-1), this statement will never succeed. while being  -1.

avpatel: hart->state being enum is treated as signed integer so "hart->start < 0" is true when hart->state = -1

slingappa: but in qemu its not behaving so. And one of [SO answer quoting ISO/IEC](https://stackoverflow.com/questions/159034/are-c-enums-signed-or-unsigned/159308#159308) also mentioning it could ?

avpatel: If thats the case then use rpmi_int32_t instead of rpmi_uint32_t

slingappa: my bad, its a typo.

PR #5, opened 2024-07-04
lib: add memory access functions to env header

slingappa:
in environments where one of the memcpy address argumnets can be a hardware address, so replaced the calls to memcpy with mem_read/write() calls where environment specific memory red/write operations can be implemented.

rpmi_shmem_simple_read/write() routines now use rpmi_env_mem_read/write() instead of memcpy().

### DIFF
--- a/include/librpmi_env.h
+++ b/include/librpmi_env.h
@@ -81,6 +81,27 @@ static inline void *rpmi_env_memcpy(void *dest, const void *src, rpmi_size_t n)
 }
 
 /**
+ * @brief Write a hw memory range with buffer
+ *
+ * @param[in] dest 	destination hw address
+ * @param[in] buf 	buffer pinter
+ * @param[in] n 	number of bytes to write
+ * @return void *	Pointer to destination
+ */
+void *rpmi_env_mem_write(char *dest, const void *buf, rpmi_size_t n);
+
+
+/**
+ * @brief read a hw memory range to buffer
+ *
+ * @param[in] src 	srour hw address
+ * @param[in] buf 	buffer pinter
+ * @param[in] n 	number of bytes to write
+ * @return void *	Pointer to destination
+ */
+void *rpmi_env_mem_read(const void *src, char* buf, rpmi_size_t n);
+
+/**
  * @brief Write or fill a memory range with a character
  *
  * @param[in] dest 	pointer to destination
@@ -94,7 +115,7 @@ static inline void *rpmi_env_memset(void *dest, int c, rpmi_size_t n)
 
 	while (n > 0) {
 		n--;
-		*temp++ = c;
+		rpmi_env_mem_write(temp++, &c, 1);
 	}
 
 	return dest;


include/librpmi_env.h
```diff
@@ -80,6 +80,27 @@ static inline void *rpmi_env_memcpy(void *dest, const void *src, rpmi_size_t n)
 	return dest;
 }
 
+/**
+ * @brief Write a hw memory range with buffer
+ *
+ * @param[in] dest 	destination hw address
+ * @param[in] buf 	buffer pinter
```
pathakraul, line 87: %s/pinter/pointer

include/librpmi_env.h
```diff
@@ -80,6 +80,27 @@ static inline void *rpmi_env_memcpy(void *dest, const void *src, rpmi_size_t n)
...
+ * @return void *	Pointer to destination
+ */
+void *rpmi_env_mem_write(char *dest, const void *buf, rpmi_size_t n);
+
+
+/**
+ * @brief read a hw memory range to buffer
+ *
+ * @param[in] src 	srour hw address
+ * @param[in] buf 	buffer pinter
```
pathakraul, line 98: %s/pinter/pointer

include/librpmi_env.h
pathakraul: How a "hw" address different from normal usual address which supports load/store?

include/librpmi_env.h
```diff
@@ -80,6 +80,27 @@ static inline void *rpmi_env_memcpy(void *dest, const void *src, rpmi_size_t n)
...
 
+/**
+ * @brief Write a hw memory range with buffer
+ *
+ * @param[in] dest 	destination hw address
+ * @param[in] buf 	buffer pinter
+ * @param[in] n 	number of bytes to write
+ * @return void *	Pointer to destination
+ */
+void *rpmi_env_mem_write(char *dest, const void *buf, rpmi_size_t n);
```
pathakraul, line 91: `char` can be `void` and leaving the granularity of write/read to the host firmware.

include/librpmi_env.h
```diff
@@ -94,7 +115,7 @@ static inline void *rpmi_env_memset(void *dest, int c, rpmi_size_t n)
 
 	while (n > 0) {
 		n--;
-		*temp++ = c;
+		rpmi_env_mem_write(temp++, &c, 1);
```
pathakraul, line 118: 1 should be `sizeof(int)`

avpatel:
The simple rpmi_shmem is only suitable for real firmwares whereas for QEMU / MockTests we will need custom rpmi_shmem implementation. In other words, we don't need the changes added by this PR.

pathakraul:
Can't we have default implementation of memcpy/memset and also given option for such host firmware apis like this mem_write/mem_read which host firmware may implement in more optimized way they can support. 

memcpy/memset in librpmi can check if mem_write/mem_read are implemented or not otherwise fallback to default implementation. This way these low level apis will find the way in librpmi which can be used for mocking also

avpatel:
It is expected that for special uses (such as QEMU or MockTest), the users will create their own custom rpmi_shmem instance instead of relying on simple rpmi_shmem. 

We can certainly improve existing simple rpmi_shmem so that some of the custom use of rpmi_shmem can avoid duplicating code. I will create a PR in some time.

avpatel:
Check this PR https://github.com/riscv-software-src/librpmi/pull/6

PR #8, opened 2024-07-09
librpmi fix hsm and suspend issues

slingappa:
- in rpmi_context_process_a2p_request() removed data length check as in cases like suspend types inquiry this check is invalid
- in rpmi_sysreset_do_reset(), while(1) was stopping the reboot by making qemu not releasing the Big Quemu Lock.

### DIFF
--- a/lib/rpmi_service_group_hsm.c
+++ b/lib/rpmi_service_group_hsm.c
@@ -214,8 +214,8 @@ static enum rpmi_error rpmi_hsm_sg_get_suspend_types(struct rpmi_service_group *
 
 	*response_datalen = (returned + 3) * sizeof(*resp);
 	resp[0] = rpmi_to_xe32(trans->is_be, (rpmi_uint32_t)status);
-	resp[1] = rpmi_to_xe32(trans->is_be, returned);
-	resp[2] = rpmi_to_xe32(trans->is_be, remaining);
+	resp[1] = rpmi_to_xe32(trans->is_be, remaining);
+	resp[2] = rpmi_to_xe32(trans->is_be, returned);
 
 	return RPMI_SUCCESS;
 }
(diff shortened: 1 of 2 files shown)


lib/rpmi_context.c
```diff
@@ -380,8 +380,7 @@ void rpmi_context_process_a2p_request(struct rpmi_context *cntx)
 			continue;
 
 		rpmi_env_lock(group->lock);
-		if (service && service->process_a2p_request &&
-		    rmsg->header.datalen >= service->min_a2p_request_datalen)
+		if (service && service->process_a2p_request)
```
avpatel, line 383: Why drop the check on datalen ?

lib/rpmi_service_group_hsm.c
```diff
@@ -214,8 +214,8 @@ static enum rpmi_error rpmi_hsm_sg_get_suspend_types(struct rpmi_service_group *
 
 	*response_datalen = (returned + 3) * sizeof(*resp);
 	resp[0] = rpmi_to_xe32(trans->is_be, (rpmi_uint32_t)status);
-	resp[1] = rpmi_to_xe32(trans->is_be, returned);
-	resp[2] = rpmi_to_xe32(trans->is_be, remaining);
+	resp[1] = rpmi_to_xe32(trans->is_be, remaining);
+	resp[2] = rpmi_to_xe32(trans->is_be, returned);
```
avpatel, line 218: This fix is fine.

lib/rpmi_service_group_sysreset.c
```diff
@@ -80,7 +80,6 @@ static enum rpmi_error rpmi_sysreset_do_reset(struct rpmi_service_group *group,
 	sysreset_type = rpmi_sysreset_find_type(sgrst, reset_type);
 	if (sysreset_type) {
 		sgrst->ops->do_system_reset(sgrst->ops_priv, reset_type);
-		while (1) ;
```
avpatel, line 83: Better to also have a DPRINTF() here.

lib/rpmi_context.c
slingappa: not needed anymore, opensbi was not seniong suspend type before, fixed now.

PR #11, opened 2024-07-14
lib: Implement clock infra and clock service group

pathakraul:
Implement Clock infrastructure to manage the system clocks and also implement RPMI Clock service group based on the clock infrastructure to enable RPMI clients to control the supported clocks via RPMI interface.

### DIFF
--- a/include/librpmi.h
+++ b/include/librpmi.h
@@ -163,6 +163,7 @@ enum rpmi_servicegroup_id {
 	RPMI_SRVGRP_SYSTEM_RESET = 0x00002,
 	RPMI_SRVGRP_SYSTEM_SUSPEND = 0x00003,
 	RPMI_SRVGRP_HSM = 0x00004,
+	RPMI_SRVGRP_CLOCK = 0x00007,
 	RPMI_SRVGRP_ID_MAX_COUNT,
 };
 
@@ -233,6 +234,19 @@ enum rpmi_hsm_service_id {
 	RPMI_HSM_SRV_GET_SUSPEND_TYPES = 0x07,
 	RPMI_HSM_SRV_GET_SUSPEND_INFO = 0x08,
 	RPMI_HSM_SRV_ID_MAX = 0x09,
+};
+
+/** RPMI Clock (CLK) ServiceGroup Service IDs */
+enum rpmi_clock_service_id {
+	RPMI_CLK_SRV_ENABLE_NOTIFICATION = 0x01,
+	RPMI_CLK_SRV_GET_NUM_CLOCKS = 0x02,
+	RPMI_CLK_SRV_GET_ATTRIBUTES = 0x03,
+	RPMI_CLK_SRV_GET_SUPPORTED_RATES = 0x04,
+	RPMI_CLK_SRV_SET_CONFIG = 0x05,
+	RPMI_CLK_SRV_GET_CONFIG = 0x06,
+	RPMI_CLK_SRV_SET_RATE = 0x07,
+	RPMI_CLK_SRV_GET_RATE = 0x08,
+	RPMI_CLK_SRV_ID_MAX,
 };
 
 /** @} */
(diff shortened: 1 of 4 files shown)


include/librpmi.h
```diff
@@ -783,6 +796,230 @@ struct rpmi_service_group {
...
+	RPMI_CLK_TYPE_DISCRETE = 0,
+	RPMI_CLK_TYPE_LINEAR = 1,
+	RPMI_CLK_TYPE_MAX_IDX,
+};
+
+/** A clock rate representation in RPMI */
+struct rpmi_clock_rate {
+	rpmi_uint32_t lo;
+	rpmi_uint32_t hi;
+};
```
avpatel, line 830: rpmi_clock_rate_match , rpmi_clock_state, rpmi_clock_type , and rpmi_clock_rate should be just after rpmi_clock_service_id  so that it is part of RPMI_MSG_PROTOCOL doxygen group.

include/librpmi.h
```diff
@@ -783,6 +796,230 @@ struct rpmi_service_group {
...
+	/* Clock rate format type */
+	enum rpmi_clock_type clock_type;
+	/* Clock name */
+	const char *name;
+	/* Clock rate array() */
+	const rpmi_uint64_t *clock_rate_array;
+};
+
+/** Clock attributes */
+struct rpmi_clock_attrs {
```
avpatel, line 855: Even rpmi_clock_attrs  should be part of RPMI_MSG_PROTOCOL doxygen group.

include/librpmi.h
```diff
@@ -235,6 +236,18 @@ enum rpmi_hsm_service_id {
...
+	RPMI_CLK_SRV_ENABLE_NOTIFICATION = 0x01,
+	RPMI_CLK_SRV_GET_NUM_CLOCKS = 0x02,
+	RPMI_CLK_SRV_GET_ATTRIBUTES = 0x03,
+	RPMI_CLK_SRV_GET_SUPPORTED_RATES = 0x04,
+	RPMI_CLK_SRV_SET_CONFIG = 0x05,
+	RPMI_CLK_SRV_GET_CONFIG = 0x06,
+	RPMI_CLK_SRV_SET_RATE = 0x07,
+	RPMI_CLK_SRV_GET_RATE = 0x08,
+	RPMI_CLK_SRV_ID_MAX,
+};
```
avpatel, line 250: Empty newline needed after this.

include/librpmi.h
```diff
@@ -783,6 +796,230 @@ struct rpmi_service_group {
 	void *priv;
 };
 
+struct rpmi_clock_module;
```
avpatel, line 799: I don't see any use-case of having a separate clock module instance.

include/librpmi.h
```diff
@@ -783,6 +796,230 @@ struct rpmi_service_group {
...
+					const struct rpmi_clock_data *clock_tree_data,
+					const struct rpmi_clock_platform_ops *ops,
+					void *ops_priv);
+
+/**
+ * @brief Destroy(free) a clock module instance
+ *
+ * @param[in] clock_mod		pointer to clock module instance
+ */
+void rpmi_clock_module_destroy(struct rpmi_clock_module *clock_mod);
```
avpatel, line 1021: We can merge rpmi_clock_module_create() and rpmi_clock_module_destroy() into rpmi_service_group_clock_create() and rpmi_service_group_clock_destroy().

include/librpmi.h
```diff
@@ -783,6 +796,230 @@ struct rpmi_service_group {
...
+
+/**
+ * @brief Get clock rate
+ *
+ * @param[in] clock_mod		pointer to clock_module instance
+ * @param[in] clock_id		clock id
+ * @param[in] rate		pointer to rate buffer
+ * @return enum rpmi_error
+ */
+enum rpmi_error rpmi_clock_get_rate(struct rpmi_clock_module *clock_mod,
```
avpatel, line 980: I don't understand why we need to expose rpmi_clock_get_clock_count(), rpmi_clock_get_attrs(), rpmi_clock_set_state(), rpmi_clock_set_rate(), and rpmi_clock_get_rate() as public APIs.

These are supposed to be internal functions of librpmi.

include/librpmi_env.h
```diff
@@ -100,6 +100,46 @@ static inline void *rpmi_env_memset(void *dest, int c, rpmi_size_t n)
...
+char *rpmi_env_strncpy(char *dest, const char *src, rpmi_size_t count)
+{
+	char *ret = dest;
+
+	while (count && *src != '\0') {
+		*dest++ = *src++;
+		count -= 1;
+	}
+
+	rpmi_env_memset(dest, '\0', count);
```
avpatel, line 139: This is not correct behaviour of strncpy(). We only need the following:

```
if (count)
    *dest = '\0';
```

avpatel:
In general, we should have minimal stuff in the public librpmi.h because thats our ABI for all users.

We have rpmi_hsm in the librpmi.h because multiple service groups namely HSM, SystemSuspend, and CPPC need a common rpmi_hsm instance.

Some of the clocks instances might be shared across different platform operations but I really doubt we need expose entire clock module instance s public ABI.

include/librpmi.h
pathakraul: I will merge the rpmi_clock.c with *service_group_clock.c. so these will go away from header

include/librpmi_env.h
pathakraul: man page says otherwise
" If the source has too few non-null bytes to fill the destination, the functions pad the destination with trailing null bytes"

include/librpmi.h
pathakraul: I made them separate to provide interface apart from service group directly to host fw in case. Since we are unsure of all the usecases currently

include/librpmi_env.h
avpatel: okay

include/librpmi.h
avpatel: In the future, we can selectively expose APIs in librpmi based on use-case.

include/librpmi.h
avpatel: @pathakraul check this comment.

avpatel:
@pathakraul Two minor comments were missed.

lib/objects.mk
```diff
@@ -9,6 +9,8 @@ lib-objs-y += rpmi_hsm.o
 lib-objs-y += rpmi_service_group_hsm.o
 lib-objs-y += rpmi_service_group_sysreset.o
 lib-objs-y += rpmi_service_group_syssusp.o
+#lib-objs-y += rpmi_clock.o
```
avpatel, line 12: Remove this line.

include/librpmi.h
pathakraul: I didnt miss but checked the state of our pdf generated and i think whole librpmi.h will need changes from doxygen perspective so i didnt do anything here

pathakraul: [refman.pdf](https://github.com/user-attachments/files/16247604/refman.pdf)

pathakraul: I will make the changes for complete librpmi.h and raise a new PR

lib/objects.mk
avpatel: @pathakraul one last comment here.

pathakraul: yes, missed this

PR #14, opened 2024-07-19
clock fixes and changes to undef all macros which librpmi headers are defining

pathakraul:


lib/librpmi_internal.h
```diff
@@ -34,20 +34,33 @@
 #define offsetof(TYPE, MEMBER) ((rpmi_size_t) &((TYPE *)0)->MEMBER)
 #endif
 
+#undef container_of
```
avpatel, line 37: Instead of `#undef` use `#ifndef` here and everywhere below.

pathakraul: We should not rely on host firmware for these because we will have no control over how these macros will be implemented by host fw

avpatel: in which case, add "rpmi_" prefix to these macro names and drop the `#undef`

pathakraul: some macros here may have common understanding but others like CLAMP, ROUNDUP, ROUNDDOWN, there is none

avpatel: adding "rpmi_" prefix does'nt change the understanding of macros.

pathakraul: that comment was before yours updated :)

pathakraul: done

include/librpmi.h
```diff
@@ -378,6 +378,7 @@ void rpmi_shmem_destroy(struct rpmi_shmem *shmem);
  * a built-in service group of the RPMI context and is always available.
  */
 struct rpmi_context;
+struct rpmi_transport;
```
avpatel, line 381: Why this change ?

pathakraul: Due to movement of the groups in header, this structure is gone below where it was referred. So instead of moving that group again above i declared here. Not sure why it didnt failed when i build with previous PR but it failed now so made the change.

avpatel: If this is to fix compile warning then move LIBRPMI_TRANSPORT_INTERFACE  before LIBRPMI_CONTEXT_INTERFACE

pathakraul: Either this or moving the group again to satisfy the symbol visibility is fine.

pathakraul: ok

PR #17, opened 2024-07-22
Add static inline to header file function declarations

slingappa:
Add static inline to header file function declarations


tests/Makefile
avpatel, line 1: This overly simplistic Makefile. Lot of things are missing such as source-level dependency tracking, multiple test applications, etc.

Let me create a separate PR which only adds the build infrastructure.

slingappa:
kept only static declaration changes.

PR #18, opened 2024-07-22
Build infrastructure for test applications

avpatel:


avpatel:
@slingappa check this whenever you can.

PR #19, opened 2024-07-22
tests: Add added base RPMI tests

slingappa:
Added Base service group tests.
Refer to test/README.md for mode details

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,7 +99,7 @@ CPPFLAGS	+=	$(GENFLAGS)
 
 ARFLAGS		=	rcs
 
-ELFFLAGS	+=	-static
+ELFFLAGS	+=	-static -L$(build_dir) -lrpmi
 ELFFLAGS	+=	$(EXTRA_ELFFLAGS)
 
 # Setup functions for compilation
@@ -140,7 +140,7 @@ compile_cc = $(CMD_PREFIX)mkdir -p `dirname $(1)`; \
 	     $(CC) $(CFLAGS) $(call dynamic_flags,$(1),$(2)) -c $(2) -o $(1)
 compile_elf = $(CMD_PREFIX)mkdir -p `dirname $(1)`; \
 	     echo " ELF       $(subst $(build_dir)/,,$(1))"; \
-	     $(CC) $(CFLAGS) $(ELFFLAGS) $(2) $(3) $(4) -o $(1)
+	     $(CC) $(CFLAGS) $(2) $(3) $(4) $(ELFFLAGS) -o $(1)
 compile_ar = $(CMD_PREFIX)mkdir -p `dirname $(1)`; \
 	     echo " AR        $(subst $(build_dir)/,,$(1))"; \
 	     $(AR) $(ARFLAGS) $(1) $(2)
--- a/include/librpmi_env.h
+++ b/include/librpmi_env.h
@@ -107,7 +107,7 @@ static inline void *rpmi_env_memset(void *dest, int c, rpmi_size_t n)
  * @param[in] count	number of max chars to count
  * @return rpmi_size_t  number of chars in string limited by count
  */
-rpmi_size_t rpmi_env_strnlen(const char *str, rpmi_size_t count)
+static inline rpmi_size_t rpmi_env_strnlen(const char *str, rpmi_size_t count)
 {
 	unsigned long ret = 0;
 
@@ -127,7 +127,7 @@ rpmi_size_t rpmi_env_strnlen(const char *str, rpmi_size_t count)
  * @param[in] count	number of max chars to count
  * @return char *	pointer to string destination buffer
  */
-char *rpmi_env_strncpy(char *dest, const char *src, rpmi_size_t count)
+static inline char *rpmi_env_strncpy(char *dest, const char *src, rpmi_size_t count)
 {
 	char *ret = dest;
 
--- a/include/librpmi_env.h
+++ b/include/librpmi_env.h
@@ -60,6 +60,32 @@ typedef long			rpmi_ssize_t;
  */
 
 /**
+ * @brief compare memory from one place to another
+ *
+ * @param[in] s1	pointer to src1
+ * @param[in] s2	pointer to src2
+ * @param[in] n		number of bytes to compare
+ * @return void *	Pointer to destination
+ */
+static inline int rpmi_env_memcmp(void *s1, void *s2, rpmi_size_t n)
+{
+	char *temp1 = s1;
+	char *temp2 = s2;
+
+	while (n > 0) {
+		if (*temp1 < *temp2)
+			return -1;
+		if (*temp1 > *temp2)
+			return 1;
+		temp1++;
+		temp2++;
+		n--;
+	}
+
+	return 0;
+}
+
+/**
  * @brief Copy memory from one place to another
  *
  * @param[in] dest 	pointer to destination
--- a/test/README.md
+++ b/test/README.md
@@ -1,0 +1,32 @@
+
+# 1. Overview
+
+This repo add RPMI tests for rpmi framework
+
+# 2. Running tests
+
+build and run tests
+
+```
+# build the tests
+$ make
+
+# execute the tests
+$ ./build/test/test_base.elf
+
+Test Base Service Group
+*******************************************
+
+Executing RPMI_BASE service group tests :
+-------------------------------------------------
+RPMI_BASE_SRV_ENABLE_NOTIFICATION                        : Succeeded!
+RPMI_BASE_SRV_PROBE_SERVICE_GROUP                        : Succeeded!
+RPMI_BASE_SRV_GET_IMPLEMENTATION_VERSION                 : Succeeded!
+RPMI_BASE_SRV_GET_SPEC_VERSION                           : Succeeded!
+RPMI_BASE_SRV_GET_IMPLEMENTATION_IDN                     : Succeeded!
+RPMI_BASE_SRV_GET_HW_INFO                                : Succeeded!
+RPMI_BASE_SRV_GET_ATTRIBUTES                             : Succeeded!
+RPMI_BASE_SRV_SET_MSI                                    : Succeeded!
+
+```
+
--- a/test/test_common.c
+++ b/test/test_common.c
@@ -5,3 +5,268 @@
  */
 
 #include "test_common.h"
+
+#define VENDOR_ID 0x1337
+#define VENDOR_SUB_ID 0x7331
+
+#define NUM_SOCKETS     1
+
+char *rpmi_shm = NULL;
+
+/* dump buffer in hexadecimal format */
+void hexdump(char *desc, unsigned int *buf, unsigned int len)
+{
+	unsigned int i;
+
+	printf("Dumping %s at %p\n", desc, buf);
+	for (i = 0; i < len/4; i++)
+		printf("%d: 0x%08x\n", i, *(buf+i));
+
+}
+
+void *rpmi_env_zalloc(rpmi_size_t size)
+{
+	return calloc(1, size);
+}
+
+void rpmi_env_free(void *ptr)
+{
+	free(ptr);
+}
+
+void rpmi_env_writel(rpmi_uint64_t addr, rpmi_uint32_t val)
+{
+	*((rpmi_uint32_t *) addr) = val;
+}
+
+struct rpmi_message *get_rpmi_msg(struct rpmi_transport *rpmi_xport)
+{
+	return rpmi_env_zalloc(rpmi_xport->slot_size);
+}
+
+void free_rpmi_msg(struct rpmi_message *msg)
+{
+	if (msg)
+		rpmi_env_free(msg);
+}
+
+void *get_shm_addr(unsigned int *shm_sz)
+{
+	if (!rpmi_shm) {
+		rpmi_shm = rpmi_env_zalloc(RPMI_SHM_SZ);
+	}
+
+	if (rpmi_shm)
+	{
+		*shm_sz = RPMI_SHM_SZ;
+	}
+
+	return rpmi_shm;
+}
+
+struct rpmi_transport *init_test_rpmi_xport(int do_clear)
+{
+	char *shm = NULL;
+	unsigned int shm_sz;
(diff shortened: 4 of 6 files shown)


Makefile
```diff
@@ -102,6 +102,8 @@ ARFLAGS		=	rcs
 ELFFLAGS	+=	-static
 ELFFLAGS	+=	$(EXTRA_ELFFLAGS)
 
+LDFLAGS	+= -L$(build_dir) -lrpmi
```
avpatel, line 105: Add these to `ELFFLAGS` just after `-static`

include/librpmi_env.h
```diff
@@ -59,6 +59,32 @@ typedef long			rpmi_ssize_t;
...
 
+/**
+ * @brief compare memory from one place to another
+ *
+ * @param[in] s1	pointer to src1
+ * @param[in] s2	pointer to src2
+ * @param[in] n		number of bytes to compare
+ * @return void *	Pointer to destination
+ */
+static inline int rpmi_env_memcmp(void *s1, void *s2, rpmi_size_t n)
```
avpatel, line 70: The change in librpmi_env.h should be a separate patch.

Makefile
avpatel: Also, the top-level Makefile change should be a separate patch.

test/test_shmem.c
```diff
@@ -0,0 +1,155 @@
...
+	/** Pointer to platform specific operations */
+	const struct rpmi_shmem_platform_ops *ops;
+	/** Private data of the platform specific oprations */
+	void *ops_priv;
+};
+
+static enum rpmi_error shmem_env_memcpy_read(void *priv, rpmi_uint64_t addr,
+					     void *in, rpmi_uint32_t len)
+{
+	rpmi_env_memcpy(in, (void *)addr, len);
```
avpatel, line 36: Never directly case integer to pointer. Instead first cast it to `unsigned long` and then cast it to pointer.

For example: `rpmi_env_memcpy(in, (void *)(unsigned long)addr, len);`

avpatel: This is to avoid compilation errors for 32bit systems.

Makefile
slingappa: Done

test/test_shmem.c
slingappa: This file is gone now.

slingappa:
@avpatel took care of the comments, please review

test/README.md
```diff
@@ -0,0 +1,74 @@
+
+# 1. Overview
+
+This repo add RPMI tests for rpmi framework implemented on quemu
+
+# 2. Running tests
+
+## 2.1 Running qemu
+
+Execute qemu instance with shared memory backend for RAM:
```
avpatel, line 10: These tests are mock test which we can directly run as application on x86 host so no need to launch QEMU.

slingappa: can be yes, but some has to write the server if they're not using the qemu right ?

avpatel:
Change the patch subject of last commit from `tests: Add added base RPMI tests` to `test: Add base RPMI tests`

test/README.md
avpatel: better to have both client and server in the same test

avpatel: Basically each test should define the messages exchanged client and server. The test fails if expected response is not received a particular message.

avpatel: One test can be something like this:

1. Enqueue a request (message)
2. Process the context to get response
3. Dequeue the response (message)
4. Inspect the response to declare pass/fail

pathakraul: If possible, a summary of all tests passed/failed too

slingappa: These are already done:
- Basically each test should define the messages exchanged client and server. The test fails if expected response is not received a particular message.

- a summary of all tests passed/failed too, in the end , please refer to README.md inside test folder for sample output

slingappa:
> test: Add base RPMI tests

done

test/README.md
slingappa: > better to have both client and server in the same test

Server code inside librpmi would bloat it, as we'll have to paste all of the qemu RPMI implentation in the server code. better to point them to upstream qemu branch which already has all of it and provide samepl qemu command , like we have in the README file no ?

slingappa: > One test can be something like this:
> 
> 1. Enqueue a request (message)
> 2. Process the context to get response
> 3. Dequeue the response (message)
> 4. Inspect the response to declare pass/fail

This is done in _base* file, please check

avpatel: Okay, sounds good.

But, we still don't need to launch QEMU for running librpmi test applications.

slingappa: We do, for now base group might work with small snippet to initialize that, but for other groups like HSM, CPPC and all still needs to form the response another server app still needs to emulate, which essentially is being done in qemu.

If we dont want to do that, then, we'll have to do one of these:
1. Use the running qemu instance
2. write a test server app inside qemu code which links all the rpmi code
3. write a test server app out side qemu copy all the rpmi code.
So I still think option#1 above seems simplest one from users point of view.

Let me know if you guys would like to implement any one of teh above still or any other server emulation, will do it.

pathakraul: I thought that our current shmem(transport) can be entirely mocked on the x86 host? 
Both sides of transport (server and client) can be invoked in sequential manner to make one complete message request response transaction. Bigger test functions can be written which will comprise multiple such function calls in sequence of both sides(server and client)

I am just thinking from a user perspective where i dont want to have qemu launch which will require me to first build it and also find the right repo and things which go with it.

Thoughts?

slingappa: > I thought that our current shmem(transport) can be entirely mocked on the x86 host?
> 
> Both sides of transport (server and client) can be invoked in sequential manner to make one complete message request response transaction. Bigger test functions can be written which will comprise multiple such function calls in sequence of both sides(server and client)
> 
> I am just thinking from a user perspective where i dont want to have qemu launch which will require me to first build it and also find the right repo and things which go with it.
> 
> Thoughts?

ofcourse, it can be mocked, as mentioned in options 2 or 3 [here](https://github.com/riscv-software-src/librpmi/pull/19#discussion_r1690923413). 

But.. see my concerns mentioned in the same comment. if we really want to we can copy qemu implementation for current and future service groups. I personally dont think its a good idea to push an test implementation which bloats up librpmi.

slingappa:
@avpatel @pathakraul ,
Added two more commits, which implements a test server to test Base group.

Now, no need to qemu, librpmi can be (mock)tested on any linux machine.

Again I'll let you guys to decide, 
- if and when we do decide this is what we need we can pick all these 5 commits,
- if we want to roll with qemu, then omit last 2 commits.

test/objects.mk
```diff
@@ -3,13 +3,22 @@
 #
 # Copyright (c) 2024 Ventana Micro Systems Inc.
 #
+#
+test-elfs-y += test_srv
```
avpatel, line 7: I still don't understand. Why separate server application ?

With this approach, each test scenario will require a separate test application.

test/test_common.c
```diff
@@ -5,3 +5,148 @@
...
+			hexdump("received",
+				(unsigned int *)msg->data,
+				rpmi_test->resp_data_len);
+			failed = 1;
+		}
+	}
+	printf("%-50s \t : %s!\n", rpmi_test->name, failed?"Failed":"Succeeded");
+}
+
+void execute_rpmi_test_group(struct rpmi_transport *rpmi_xport, struct rpmi_test_group *group)
```
avpatel, line 77: We can improve this common test infrastructure such that each test has both client and server part of the test-case. This will help us avoid creating a separate server application for each test scenario.

test/objects.mk
slingappa: No, it doesnt, 

` process_ap_reqs()` inside `test_srv.c ` is common for all groups,
but server implementation definitely needs to be added just like we did in qemu for each service groups seperately as depicted below in  `add_*_group()` below:

```
 int main(int argc, char *argv[])
 {
         printf("Test server thread...\n");

         /*
          * along with contexts only BASE group is created, rest needs to be
          * added here later, after the context creation
          */
         create_contexts();
         
         add_hsm_group(&hsm);
         add_cppc_group(&cppc)
         
         process_ap_reqs();

         return 0;
 }

```

avpatel: This approach of adding a separate server for various scenarios is not going to scale. We have to re-design the common test infrastructure.

test/test_common.h
```diff
@@ -7,4 +7,90 @@
...
+	int (*init)(struct rpmi_test *test);
+	int (*run)(struct rpmi_test *test);
+	int (*wait)(struct rpmi_test *test, struct rpmi_message *msg);
+	int (*verify)(struct rpmi_test *test, struct rpmi_message *msg);
+	int (*cleanup)(struct rpmi_test *test);
+};
+
+struct rpmi_test_group {
+	char name[64];
+	int num_tests;
```
avpatel, line 70: additonal members

rpmi_shmem *shmem;
rpmi_transport *trans;
rpmi_context *context;

test/test_common.h
```diff
@@ -7,4 +7,90 @@
...
+	struct rpmi_test tests[];
+};
```
avpatel, line 72: setup()
cleanup()

slingappa:
@avpatel @pathakraul ,

please check the changes now modified as per our discussion.

users are expected to modify `struct rpmi_test_group` members to create different test scenarios. and run `execute_test_group` repeatedly.

test/test_common.h
```diff
@@ -7,4 +7,126 @@
...
+#include <stdio.h>
+#include <librpmi.h>
+#include <stdlib.h>
+
+/*
+ * Qemu shared memory offsets.
+ * its better to decode these data from DTS file, hardcoding for now.
+ */
+
+#define RPMI_SHM_XPORTS       (2)
```
avpatel, line 19: Drop this.

test/test_common.h
```diff
@@ -7,4 +7,126 @@
...
+/*
+ * test assums qemu RPMI shared memory address is mapped at the following
+ * offset in gues RAM
+ */
+#define RPMI_SHM_OFSET       (0x0000000010001000)
```
avpatel, line 24: We don't need lot of these defines.

test/test_common.h
```diff
@@ -7,4 +7,126 @@
...
+		rpmi_uint32_t hw_info_len;
+		const rpmi_uint8_t *hw_info;
+	} base;
+	char *priv;
+	int priv_len;
+};
+
+struct rpmi_test {
+	char name[64];
+	void *rpmi_xport;
```
avpatel, line 71: No need for this pointer. See below.

test/test_common.h
```diff
@@ -7,4 +7,126 @@
...
+		unsigned int svc_id;
+		unsigned int svc_type;
+	} attrs;
+	int query_data_len;
+	unsigned char query_data[RPMI_MSG_MAX_DATA_SIZE];
+	int resp_data_len;
+	unsigned char resp_data[RPMI_MSG_MAX_DATA_SIZE];
+
+	/* test specific function handlers*/
+	int (*init)(struct rpmi_test *test);
```
avpatel, line 85: For all the test callbacks, just pass rpmi_test_group pointer which already has rpmi_context and rpmi_transport pointers.

test/test_transport.c
```diff
@@ -0,0 +1,155 @@
...
+		rmsg->header.flags,
+		rmsg->header.servicegroup_id,
+		rmsg->header.service_id,
+		rmsg->header.datalen, rmsg->header.token);
+
+	for (i = 0; i < (rmsg->header.datalen / 4); i++)
+		printf("data[%2d]: %08x \n", i, data[i]);
+}
+
+void *get_shm_addr(unsigned int xport_id, unsigned int *shm_sz)
```
avpatel, line 35: We don't need this anymore.

test/test_transport.c
```diff
@@ -0,0 +1,155 @@
...
+	return calloc(1, (rpmi_xport->slot_size));
+}
+
+void free_rpmi_msg(struct rpmi_message *msg)
+{
+	if (msg)
+		free(msg);
+}
+
+struct rpmi_transport *init_test_rpmi_xport(int xport_id, int do_clear)
```
avpatel, line 114: This function is not need. Most of this things can be done in init_group().

In fact, you can move important stuff from test_transport.c and remove test_transport.c

test/test_common.c
```diff
@@ -5,3 +5,237 @@
...
+			group->init(group);
+	}
+
+	return 0;
+}
+
+void cleanup_group(struct rpmi_test_group *group)
+{
+	free(group->xport->priv);
+	free(group->rctx);
```
avpatel, line 208: Use librpmi APIs.

test/test_common.h
slingappa: agreed, dropping.

test/test_common.h
slingappa: agreed, was needed for qemu, not anymore dropping.

test/test_common.h
slingappa: No, rpmi_test still needs as we still need to track, request/response data per service group tests.

slingappa:
@avpatel 
- `test_common.c` allocates/frees for every test memory instead of opening `/dev/shm/` file
- removed `*transport.c` file and brought in required changes inside `common.c` file
- removed multi transport related changes, so `xport_id` gone from everywhere.

include/librpmi_env.h
```diff
@@ -368,7 +394,10 @@ static inline rpmi_uint32_t rpmi_to_xe32(rpmi_bool_t is_be, rpmi_uint32_t val)
  * @param[in] addr	64-bit address
  * @param[in] val	4-byte data to write
  */
-void rpmi_env_writel(rpmi_uint64_t addr, rpmi_uint32_t val);
+static inline void rpmi_env_writel(rpmi_uint64_t addr, rpmi_uint32_t val)
```
avpatel, line 397: This should be implemented in test_common.c

avpatel:
@slingappa Much better now. Few comments remaining.

test/test_common.c
slingappa: Done

PR #22, opened 2024-08-06
lib: Optimize the definition of rpmi_hsm

wxjstz:
rpmi_hsm.leaf and rmpi_hsm.noleaf can be put into a union to save memory


avpatel:
Looks good to me. I will let @pathakraul decide.

pathakraul:
Looks good to me

PR #23, opened 2024-08-16
lib: Provide simple cache-non-coherent shared memory operations

avpatel:


slingappa (approved):

slingappa (commented):
I think these changes are not nessessory, its implied that *_env() implementations should take care of these cases.

avpatel:
> I think these changes are not nessessory, its implied that *_env() implementations should take care of these cache coherency.

It is important to keep separate `*_env()` functions for cache maintenance because the `mem*_env()` functions will be used on regular memory as well as shared memory.

pathakraul:
Looks good to me

PR #27, opened 2024-09-26
test: correctly check for NULL pointer during scenario init

00xc:
After attempting to allocate an RPMI context, we should verify that such allocation succeeded, but the wrong field was being checked.


pathakraul:
Looks good to me

PR #28, opened 2024-09-27
lib: clock service group fixes

00xc:
* Fix incorrect bounds checking for clock IDs coming from the AP.
* Fix an incorrect call to `DPRINTF()`.


pathakraul:
Thanks for catching this. both patches looks good to me.

PR #29, opened 2024-09-27
lib: properly release HSM hart array

00xc:
The hart array in the `rpmi_hsm` struct is created in `rpmi_hsm_create()` via `rpmi_env_zalloc()`, so it should be freed with `rpmi_env_free()`.


pathakraul:
Looks good.

PR #30, opened 2024-10-11
librpmi: syssusp: Complete state changes

jones-drew:
Reverse the logic of the state change checks, because, when they are successful, we want to do the state change, not skip the change with an early return.

### DIFF
--- a/lib/rpmi_service_group_syssusp.c
+++ b/lib/rpmi_service_group_syssusp.c
@@ -174,7 +174,7 @@ static enum rpmi_error rpmi_syssusp_process_events(struct rpmi_service_group *gr
 	case RPMI_SYSSUSP_STATE_SUSPEND_PENDING:
 		if (!sgsusp->ops->system_suspend_ready(sgsusp->ops_priv,
 						       sgsusp->current_hart_index))
-			return RPMI_SUCCESS;
+			return RPMI_ERR_BUSY;
 		sgsusp->ops->system_suspend_finalize(sgsusp->ops_priv,
 						sgsusp->current_hart_index,
 						sgsusp->current_syssusp_type,
@@ -184,7 +184,7 @@ static enum rpmi_error rpmi_syssusp_process_events(struct rpmi_service_group *gr
 	case RPMI_SYSSUSP_STATE_SUSPENDED:
 		if (!sgsusp->ops->system_suspend_can_resume(sgsusp->ops_priv,
 							    sgsusp->current_hart_index))
-			return RPMI_SUCCESS;
+			return RPMI_ERR_BUSY;
 		status = sgsusp->ops->system_suspend_resume(sgsusp->ops_priv,
 							sgsusp->current_hart_index,
 							sgsusp->current_syssusp_type,
--- a/lib/rpmi_context.c
+++ b/lib/rpmi_context.c
@@ -458,7 +458,7 @@ void rpmi_context_process_group_events(struct rpmi_context *cntx,
 	rpmi_env_lock(group->lock);
 	rc = group->process_events(group);
 	rpmi_env_unlock(group->lock);
-	if (rc) {
+	if (rc && rc != RPMI_ERR_BUSY) {
 		DPRINTF("%s: %s: group %s failed with error %d\n",
 			__func__, cntx->name, group->name, rc);
 	}
@@ -487,7 +487,7 @@ void rpmi_context_process_all_events(struct rpmi_context *cntx)
 		rpmi_env_lock(group->lock);
 		rc = group->process_events(group);
 		rpmi_env_unlock(group->lock);
-		if (rc) {
+		if (rc && rc != RPMI_ERR_BUSY) {
 			DPRINTF("%s: %s: group %s failed with error %d\n",
 				__func__, cntx->name, group->name, rc);
 		}


lib/rpmi_service_group_syssusp.c
```diff
@@ -172,19 +172,19 @@ static enum rpmi_error rpmi_syssusp_process_events(struct rpmi_service_group *gr
...
 						sgsusp->current_resume_addr);
 		sgsusp->current_state = RPMI_SYSSUSP_STATE_SUSPENDED;
 		break;
 	case RPMI_SYSSUSP_STATE_SUSPENDED:
-		if (!sgsusp->ops->system_suspend_can_resume(sgsusp->ops_priv,
-							    sgsusp->current_hart_index))
-			return RPMI_SUCCESS;
+		if (sgsusp->ops->system_suspend_can_resume(sgsusp->ops_priv,
+							   sgsusp->current_hart_index))
+			return RPMI_ERR_BUSY;
```
pathakraul, line 187: IMO there are two things missing in librpmi handling of such case -
- A error case like `RPMI_ERR_EAGAIN` which is more appropriate for such scenario.  We don't have such in the RPMI message protocol (not yet needed) but we can have an internal error code in librpmi.
- Handling of such error case in the rpmi context which does not means the failure but a hint to the caller that it has to be tried again.

lib/rpmi_service_group_syssusp.c
```diff
@@ -172,19 +172,19 @@ static enum rpmi_error rpmi_syssusp_process_events(struct rpmi_service_group *gr
 
 	switch (sgsusp->current_state) {
 	case RPMI_SYSSUSP_STATE_SUSPEND_PENDING:
-		if (!sgsusp->ops->system_suspend_ready(sgsusp->ops_priv,
-						       sgsusp->current_hart_index))
-			return RPMI_SUCCESS;
+		if (sgsusp->ops->system_suspend_ready(sgsusp->ops_priv,
+						      sgsusp->current_hart_index))
+			return RPMI_ERR_BUSY;
```
avpatel, line 177: This looks incorrect because we should be doing nothing when system is not ready to suspend.

lib/rpmi_service_group_syssusp.c
```diff
@@ -172,19 +172,19 @@ static enum rpmi_error rpmi_syssusp_process_events(struct rpmi_service_group *gr
...
 						sgsusp->current_hart_index,
 						sgsusp->current_syssusp_type,
 						sgsusp->current_resume_addr);
 		sgsusp->current_state = RPMI_SYSSUSP_STATE_SUSPENDED;
 		break;
 	case RPMI_SYSSUSP_STATE_SUSPENDED:
-		if (!sgsusp->ops->system_suspend_can_resume(sgsusp->ops_priv,
-							    sgsusp->current_hart_index))
-			return RPMI_SUCCESS;
+		if (sgsusp->ops->system_suspend_can_resume(sgsusp->ops_priv,
```
avpatel, line 185: Same comment as above. We should be doing nothing when system can't resume.

jones-drew:
On October 12, 2024 6:56:56 AM GMT+02:00, Anup Patel ***@***.***> wrote:
>@avpatel commented on this pull request.
>
>
>
>> @@ -172,19 +172,19 @@ static enum rpmi_error rpmi_syssusp_process_events(struct rpmi_service_group *gr
> 
> 	switch (sgsusp->current_state) {
> 	case RPMI_SYSSUSP_STATE_SUSPEND_PENDING:
>-		if (!sgsusp->ops->system_suspend_ready(sgsusp->ops_priv,
>-						       sgsusp->current_hart_index))
>-			return RPMI_SUCCESS;
>+		if (sgsusp->ops->system_suspend_ready(sgsusp->ops_priv,
>+						      sgsusp->current_hart_index))
>+			return RPMI_ERR_BUSY;
>
>This looks incorrect because we should be doing nothing when system is not ready to suspend.
>

We won't do anything, except return ebusy, assuming a return of zero for the ready check means it's ready and a nonzero value means it's not.

jones-drew:
On October 12, 2024 6:57:40 AM GMT+02:00, Anup Patel ***@***.***> wrote:
>@avpatel commented on this pull request.
>
>
>
>>  		sgsusp->ops->system_suspend_finalize(sgsusp->ops_priv,
> 						sgsusp->current_hart_index,
> 						sgsusp->current_syssusp_type,
> 						sgsusp->current_resume_addr);
> 		sgsusp->current_state = RPMI_SYSSUSP_STATE_SUSPENDED;
> 		break;
> 	case RPMI_SYSSUSP_STATE_SUSPENDED:
>-		if (!sgsusp->ops->system_suspend_can_resume(sgsusp->ops_priv,
>-							    sgsusp->current_hart_index))
>-			return RPMI_SUCCESS;
>+		if (sgsusp->ops->system_suspend_can_resume(sgsusp->ops_priv,
>
>Same comment as above. We should be doing nothing when system can't resume.
>

also same reply

avpatel:
> On October 12, 2024 6:56:56 AM GMT+02:00, Anup Patel ***@***.***> wrote: @avpatel commented on this pull request. > @@ -172,19 +172,19 @@ static enum rpmi_error rpmi_syssusp_process_events(struct rpmi_service_group *gr switch (sgsusp->current_state) { case RPMI_SYSSUSP_STATE_SUSPEND_PENDING: - if (!sgsusp->ops->system_suspend_ready(sgsusp->ops_priv, - sgsusp->current_hart_index)) - return RPMI_SUCCESS; + if (sgsusp->ops->system_suspend_ready(sgsusp->ops_priv, + sgsusp->current_hart_index)) + return RPMI_ERR_BUSY; This looks incorrect because we should be doing nothing when system is not ready to suspend.
> We won't do anything, except return ebusy, assuming a return of zero for the ready check means it's ready and a nonzero value means it's not.

The `if()" check modification is wrong and does not align with `system_suspend_ready()` semantics. The concern with returning EBUSY is that when debug prints are enabled, the `rpmi_context_process_all_events()` or `rpmi_context_process_group_events()` will throw lot of prints because these functions are called in polling fashion.

pathakraul:
Since this condition is internal (no response required to AP) to librpmi where it has to wait, why cannot we define internal errors like OpenSBI has. This situation is inevitable and will occur at other places also. Whatever we return for READY and RESUME checks it has to be filtered out in the rpmi_context code since we are not returning the error in status.

pathakraul:
For the `if` condition i think this demands a AI on me to improve the doc comments for these interfaces to explain explicitly what it returns in which condition

jones-drew:
I initially missed that rpmi_syssusp_platform_ops.system_suspend_ready and rpmi_syssusp_platform_ops.system_suspend_can_resume return type rpmi_bool_t (the implementation I was looking at only returned zero, and I assumed zero was SUCCESS). So we _do_ want "if (!sgsusp->ops->system_suspend_ready(...))" and "if (!sgsusp->ops->system_suspend_can_resume(...))" in rpmi_syssusp_process_events(), but I'm not sure we want to return RPMI_SUCCESS in those cases. Something like RPMI_ERR_EAGAIN seems like it would make more sense.

avpatel:
> I initially missed that rpmi_syssusp_platform_ops.system_suspend_ready and rpmi_syssusp_platform_ops.system_suspend_can_resume return type rpmi_bool_t (the implementation I was looking at only returned zero, and I assumed zero was SUCCESS). So we _do_ want "if (!sgsusp->ops->system_suspend_ready(...))" and "if (!sgsusp->ops->system_suspend_can_resume(...))" in rpmi_syssusp_process_events(), but I'm not sure we want to return RPMI_SUCCESS in those cases. Something like RPMI_ERR_EAGAIN seems like it would make more sense.

We can return `RPMI_ERR_EAGAIN` from `rpmi_syssusp_process_events()` as long as we also update `rpmi_context_process_all_events()` and `rpmi_context_process_group_events()` to avoid a debug print for `RPMI_ERR_EAGAIN`.

pathakraul:
@jones-drew Since `RPMI_ERR_EAGAIN` has no use case as a RPMI supported Error Code which are returned in the message and will not be the part of specification. I would suggest to rename it to `RPMI_EAGAIN` with the integer value as `-1000`.  It still falls in Vendor Specific though. Renaming will help to distinguish which are possibly getting returned in the message and which are not.

jones-drew:
> @jones-drew Since `RPMI_ERR_EAGAIN` has no use case as a RPMI supported Error Code which are returned in the message and will not be the part of specification. I would suggest to rename it to `RPMI_EAGAIN` with the integer value as `-1000`. It still falls in Vendor Specific though. Renaming will help to distinguish which are possibly getting returned in the message and which are not.

I'm probably misunderstanding, but I wouldn't expect we'd want to use a vendor specific error code in this generic function. If we don't want to add some sort of EAGAIN, then I think my original proposal of EBUSY should work. Reporting a failure due to being "busy" is just another way of saying "you failed this time, since I happened to be busy at the time, but maybe you won't fail when you try again", which, to me, is the same as EAGAIN.

I've updated the PR branch. The changes are to keep the !check logic, since it should be a boolean with false meaning "not ready", and to update the callers of process-events to skip debug print for EBUSY, since EBUSY could be tolerated.

PR #31, opened 2024-10-17
librpmi: Update attribute and bits definition

yeongjoshua:
Update bits definition for clock, hsm, system
suspend and system reset according to RPMI spec
v1.0-rc1.


pathakraul:
Changes looks good.

Can you separate the change in patches per service group?

yeongjoshua:
> Changes looks good.
> 
> Can you separate the change in patches per service group?

I split the changes into different patches :)

pathakraul:
Looks good

PR #44, opened 2024-10-28
lib: Update the error codes as per the latest RPMI spec

pathakraul:
RPMI specification has updated the possible error codes in the message protocol. Update the error codes across the service groups as per the latest spec.


lftan:
Changes looks okay.

PR #49, opened 2024-11-21
lib: Return service group version while probing a group

pathakraul:
Probing a service group returns the service group version in response message. Add version for each service group currently matching the RPMI spec version and return it in the response if the group is present in an context

### DIFF
--- a/lib/rpmi_service_group_syssusp.c
+++ b/lib/rpmi_service_group_syssusp.c
@@ -244,6 +244,8 @@ rpmi_service_group_syssusp_create(struct rpmi_hsm *hsm,
 	group = &sgsusp->group;
 	group->name = "syssusp";
 	group->servicegroup_id = RPMI_SRVGRP_SYSTEM_SUSPEND;
+	group->servicegroup_version =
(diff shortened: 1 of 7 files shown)


lib/rpmi_service_group_syssusp.c
```diff
@@ -244,6 +244,8 @@ rpmi_service_group_syssusp_create(struct rpmi_hsm *hsm,
 	group = &sgsusp->group;
 	group->name = "syssusp";
 	group->servicegroup_id = RPMI_SRVGRP_SYSTEM_SUSPEND;
+	group->servicegroup_version =
+		RPMI_BASE_VERSION(RPMI_SPEC_VERSION_MAJOR, RPMI_SPEC_VERSION_MINOR);
```
lftan, line 248: For now, service group version is same with RPMI version.
But, each service group should have its version (their version macros)?

pathakraul: Yes, when they will deviate we can define the new MAJOR and MINOR macros

lftan: Okay

pathakraul: Shall we merge it then?

PR #55, opened 2025-02-27
Add build time parameter to control the librpmi debug logs

pathakraul:


Makefile
```diff
@@ -98,6 +98,11 @@ endif
 GENFLAGS	=	-Wall -Werror -g -O2
 GENFLAGS	+=	-I$(include_dir) -I$(lib_dir)
 
+ifeq ($(LIBRPMI_LOG_DEBUG),y)
```
avpatel, line 101: Why not just call it LIBRPMI_DEBUG ?

Makefile
```diff
@@ -98,6 +98,11 @@ endif
...
+GENFLAGS        +=         -DDEBUG
+endif
+
+
```
avpatel, line 105: Redundant newline.

Makefile
pathakraul: LIBRPMI_DEBUG can have another meaning -> enable the debug symbols (-g) which is on by default. So keeping that would have given wrong meaning.

Initially I thought to enable DEBUG logs and also the `-g` flag together from LIBRPMI_DEBUG only but keeping `-g` on by default may be useful.

Makefile
pathakraul: Updating

Makefile
pathakraul: Thats why LIBRPMI_LOG_DEBUG was next choice for this

pathakraul: I have changed to LIBRPMI_DEBUG and also now turning off compiler optimizations with enabling logs to make the macro name suitable.

avpatel:
Looks good to me.

PR #56, opened 2025-03-11
Remove the default shmem platform operations from shmem core and let platform implement them

pathakraul:


avpatel:
If a platform has non-coherent shared memory and requires explicit cache flushing then PuC firmware should be able to do it using non-coherent operations and this does not violate RPMI spec because platform is trying to work-around coherency in platform specific way. Also, we can use memcpy() functions for internal data movement as well, platform should not blindly do cache flushing in memcpy() functions.

The simple shared memory operations are quite useful because platforms with coherent shared memory can just re-use these common operations.

My suggestion is to not go ahead with the PR.

pathakraul:
sure, will close the MR

PR #57, opened 2025-05-06
lib: Remove p2a_msi_index from rpmi_sysmsi_get_attrs() response

avpatel:
The p2a doorbell system MSI index is no longer part of the rpmi_sysmsi_get_attrs() response as-per the latest frozen RPMI specification hence remove it.


pathakraul:
Looks good

PR #59, opened 2025-08-10
lib: clock: fix potential NULL pointer dereference

00xc:
Check if allocation failed before dereferencing the clock tree pointer. The caller of rpmi_clock_tree_init() will already do the right thing if we return a NULL pointer.


pathakraul:
Looks good to me, thanks for fixing this.

PR #60, opened 2025-08-11
lib: sysmsi: fix NULL pointer check

00xc:
Fix a NULL check after attempting to allocate the system MSI array. The wrong check could cause a failed allocation to result in a NULL pointer dereference down the line.


pathakraul:
Looks good to me, thanks for fixing this

PR #78, opened 2026-04-22
lib: Add Logging service group

slingappa:
This service group defines services used by an application processor to set the log data defined by the UEFI Platform Initialization specification or other similar implementations. This service group provides a reliable way to track the progress of the system for diagnostic purposes. The data provided by the services can then be relayed to other consumers in the system such as the BMC, LCD display, etc.

### DIFF
--- a/include/librpmi.h
+++ b/include/librpmi.h
@@ -198,6 +198,7 @@ enum rpmi_servicegroup_id {
 	RPMI_SRVGRP_MANAGEMENT_MODE	= 0x000B,
 	RPMI_SRVGRP_RAS_AGENT		= 0x000C,
 	RPMI_SRVGRP_REQUEST_FORWARD	= 0x000D,
+	RPMI_SRVGRP_LOGGING			= 0x000E,
 	RPMI_SRVGRP_ID_MAX_COUNT,
 
 	/* Reserved range for service groups */
@@ -367,6 +368,13 @@ enum rpmi_mm_service_id {
 	RPMI_MM_SRV_GET_ATTRIBUTES	= 0x02,
 	RPMI_MM_SRV_COMMUNICATE		= 0x03,
 	RPMI_MM_SRV_ID_MAX
+};
+
+/** RPMI LOGGING ServiceGroup Service IDs */
+enum rpmi_logging_service_id {
+	RPMI_LOGGING_SRV_ENABLE_NOTIFICATION = 0x01,
+	RPMI_LOGGING_SRV_SET_CONFIG = 0x02,
+	RPMI_LOGGING_SRV_ID_MAX,
 };
 
 /** @} */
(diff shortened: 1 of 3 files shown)


include/librpmi.h
```diff
@@ -369,6 +370,13 @@ enum rpmi_mm_service_id {
 	RPMI_MM_SRV_ID_MAX
 };
```
pathakraul, line 372: Create a `\defgroup` for this service group. Check other groups for reference in this header

lib/rpmi_service_group_logging.c
```diff
@@ -0,0 +1,115 @@
+/*
```
pathakraul, line 2: Seems this whole file indentation is using spaces. Please use checkpatch.pl from linux/scripts to check and resolve formatting issues

lib/rpmi_service_group_logging.c
```diff
@@ -0,0 +1,115 @@
...
+        struct rpmi_service_group group;
+};
+
+static enum rpmi_error
+rpmi_logging_set_state(struct rpmi_logging_group *sglogging,
+                       void *data, rpmi_uint8_t is_be)
+{
+        enum rpmi_error ret = 0;
+
+        sglogging->ops->do_set_logging(sglogging->ops_priv, data, is_be);
```
pathakraul: What if `ops != NULL` but `do_set_logging == NULL`.

You can check this function pointer explicitly before creating the service group in function `rpmi_service_group_logging_create`

lib/rpmi_service_group_logging.c
```diff
@@ -0,0 +1,115 @@
...
+
+        return ret;
```
pathakraul: Its always 0

pathakraul: IMO the callback `do_set_logging` must return error instead of `void` because you cannot know what kind of issues may occur in platform

include/librpmi.h
```diff
@@ -2098,6 +2106,29 @@ enum rpmi_error rpmi_mm_service_register(struct rpmi_service_group *group,
 					 rpmi_uint32_t num_entries,
 					 struct rpmi_mm_service *iplist);
 
+/** Platform specific logging operations */
+struct rpmi_logging_platform_ops {
+	void (*do_set_logging)(void *priv, void *data, rpmi_uint8_t is_be);
```
pathakraul: NIT: `do_set_state` instead of `do_set_logging`

pathakraul: How the implementation of callback will know the length of the `data`. Service already gives you this in `request_datalen`

lib/rpmi_service_group_logging.c
```diff
@@ -0,0 +1,115 @@
...
+                         const rpmi_uint8_t *request_data,
+                         rpmi_uint16_t *response_datalen,
+                         rpmi_uint8_t *response_data)
+{
+        enum rpmi_error status;
+        struct rpmi_logging_group *logginggrp = group->priv;
+        rpmi_uint32_t *resp = (void *)response_data;
+
+        /* change logging config synchronously */
+        status = rpmi_logging_set_state(logginggrp, (void *)request_data,
```
pathakraul: Any reason endianness handling of `request_data` is passed on to the callback implementation instead of handling it here like in other service groups??

pathakraul: IMO callback must be kept simple and `request_data` endianness must be taken care here in this function only. Also the `request_data` must be unpacked and callback parameters can be 
`do_set_logging(void *priv, uint32_t log_type, uint32_t datalen_bytes, void *data)`

pathakraul:
Change your mailid to oss one in this PR also

include/librpmi.h
slingappa: Added

lib/rpmi_service_group_logging.c
slingappa: fixed

lib/rpmi_service_group_logging.c
slingappa: done

lib/rpmi_service_group_logging.c
slingappa: fixed as per suggestion

include/librpmi.h
slingappa: fixed

slingappa:
fixed and enhanced.

slingappa:
@pathakraul  **BUMP**

pathakraul:
@slingappa I cannot see changes for multiple comments which have marked resolved

slingappa:
@pathakraul  **BUMP**

lib/rpmi_service_group_logging.c
```diff
@@ -0,0 +1,131 @@
...
+#include <librpmi.h>
+
+#ifdef DEBUG
+#define DPRINTF(msg...)	rpmi_env_printf(msg)
+#else
+#define DPRINTF(msg...)
+#endif
+
+struct rpmi_logging_group {
+	struct rpmi_logging *logging;
```
pathakraul: Where is `struct rpmi_logging` defined. Possibly stray field

slingappa: it was, removed.

PR #80, opened 2026-05-06
lib: voltage: fix potential deadlocks

00xc:
The voltage service group mismanages locks in several ways.

First, when querying or modifying a voltage domain it acquires that domain's lock, and then re-acquires it in the corresponding double-underscore function (e.g. `rpmi_volt_get_config()` and then `__rpmi_volt_get_config()`). If a library user provides a non-recursive lock (e.g. `PTHREAD_MUTEX_NORMAL`) this will cause a deadlock.

Second, the non-underscore functions skip unlocking in multiple places if the double-underscore function fails by using a goto. This leaves the lock in an inconsistent state, likely causing a deadlock as well.

Remove all the redundant locking, as the lower level functions already do it and properly manage the unlocking even if there is a failure.


pathakraul:
Good catch. Looks good to me

PR #82, opened 2026-05-13
build: add shared librpmi and pkg-config metadata

slingappa:
Build and install versioned shared-library artifacts (librpmi.so.<version>, librpmi.so.<soversion>, librpmi.so), generate librpmi.pc, and update clean/install rules accordingly while keeping static library output.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,22 @@ $(error Install root directory is same as build directory.)
 endif
 docs_dir=$(build_dir)/docs
 install_docs_dir=$(install_root_dir)/docs
+LIBRPMI_HEADER := $(src_dir)/include/librpmi.h
+LIBRPMI_IMPL_VERSION_MAJOR := $(strip $(shell sed -n 's/^#define[[:space:]]\+LIBRPMI_IMPL_VERSION_MAJOR[[:space:]]\+\([0-9][0-9]*\).*/\1/p' $(LIBRPMI_HEADER)))
+LIBRPMI_IMPL_VERSION_MINOR := $(strip $(shell sed -n 's/^#define[[:space:]]\+LIBRPMI_IMPL_VERSION_MINOR[[:space:]]\+\([0-9][0-9]*\).*/\1/p' $(LIBRPMI_HEADER)))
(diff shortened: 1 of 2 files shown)


pathakraul:
Missing `Signed-off-by`. You are the first to violate the policy you just wrote :)

slingappa:
@pathakraul comments fixed let me know once the review is done.

Makefile
```diff
@@ -38,6 +38,9 @@ $(error Install root directory is same as build directory.)
 endif
 docs_dir=$(build_dir)/docs
 install_docs_dir=$(install_root_dir)/docs
+LIBRPMI_VERSION ?= 1.0.0
```
avpatel: Parse LIBRPMI_IMPL_VERSION_MAJOR and LIBRPMI_IMPL_VERSION_MINOR from include/librpmi.h so that version is defined in only one place which is include/librpmi.h

Makefile
```diff
@@ -161,8 +165,26 @@ compile_elf = $(CMD_PREFIX)mkdir -p `dirname $(1)`; \
...
+	       $(LDFLAGS) $(EXTRA_LDFLAGS) -o $(1) $(2)
+gen_pkgconfig = $(CMD_PREFIX)mkdir -p `dirname $(1)`; \
+	     echo " GEN       $(subst $(build_dir)/,,$(1))"; \
+	     printf '%s\n' \
+	       'prefix=/usr' \
+	       'libdir=$${prefix}/lib' \
+	       'includedir=$${prefix}/include' \
+	       '' \
+	       'Name: librpmi' \
+	       'Description: RISC-V RPMI protocol helper library' \
```
avpatel: "RISC-V Platform Management Interface (RPMI) protocol helper library"

avpatel (changes requested):

pathakraul:
Can you please also mention about the shared library objects being also generated now when `make` is called in README.md at below section

https://github.com/riscv-software-src/librpmi/blob/main/README.md#librpmia

pathakraul:
Rest looks good to me

slingappa:
ALL done

slingappa:
@avpatel  @pathakraul  **BUMP**

PR #84, opened 2026-05-13
lib: clock: allow rate change on disabled clocks

ch-perry:
Changing the rate of a disabled clock is a common thing to do and should be allowed. Otherwise, there is no way to make sure that the clock has the right frequency when it is first enabled.

This makes it possible to use any Linux driver which uses "clk_set_rate()" before "clk_prepare_enable()", for which a random example is [1].

[1]: https://elixir.bootlin.com/linux/v7.0.6/source/drivers/iio/adc/ad4130.c#L1882


slingappa:
Looks OK to me.

PR #85, opened 2026-05-14
packaging: add Debian and RPM package recipes for librpmi

slingappa:
Add distro packaging metadata and build integration for librpmi.

Debian packaging now produces:
- librpmi0 (runtime shared library)
- librpmi-dev (headers and development metadata)
- librpmi-src (source-content package)

Add an RPM spec to generate equivalent runtime, -devel, and source subpackages, and wire the packaging flow through the repository Makefile.


slingappa:
@avpatel  @pathakraul please review

pathakraul:
Why we need debian and RPM packages for this library

pathakraul:
Why `debian` folder cannot be inside the `packaging` in parallel to `rpm` folder. It will reduce the clutter at the top level

slingappa:
> Why we need debian and RPM packages for this library
This is how we can install the libraries on any host so that applications can find and link.
Something like qemu today cimfigures and links ssl or  slirp library

slingappa:
agree that debian foler can go inside. now single packaging folder has deb and rpm folders, @pathakraul

pathakraul:
Looks good to me

PR #87, opened 2026-05-18
lib: ack MM communicate errors

slingappa:
Make the Management Mode COMMUNICATE handler encode error status in the response data and return RPMI_SUCCESS to the context so normal requests still receive acknowledgements.

This fixes the unknown-GUID regression added to the MM service group test, which otherwise exposes a missing ACK for a normal request.


saini-ranbirs:
Covered by PR 90 freshly

PR #89, opened 2026-05-19
release: bump version to 1.0.0

slingappa:
Grandfathered release tag for the current librpmi baseline.
     - Version: implementation 1.0.0 from include/librpmi.h, patch 0.
     - RPMI spec: 1.0.
     - Service groups: Base, System Reset, System Suspend, HSM, Clock, CPPC, System MSI, Device Power, Performance, Voltage, MM.
     - Library: static archive plus shared library librpmi.so.1.0.0 with SONAME librpmi.so.0.
     - Packaging: Debian and RPM package release 1.0.0-0 with runtime, development, and source package infrastructure.


packaging/rpm/librpmi.spec
```diff
@@ -1,6 +1,6 @@
 Name:           librpmi
 Version:        %{?version}%{!?version:1.0.0}
-Release:        1%{?dist}
+Release:        0%{?dist}
 Summary:        RISC-V RPMI helper library
```
avpatel, line 4: RISC-V Platform Management Interface (RPMI) helper library.

PR #90, opened 2026-05-20
lib: mm: Fix request data and response data related issues

saini-ranbirs:
- Correct expected request data length for MM_COMMUNICATE call.
  - Make return to RPMI MM call as RPMI_SUCCESS always. Actual status of the call is in the response word 0.
  - Return response data length to the RPMI MM call as defined by the RPMI spec irrespective of status value in the response word 0. The caller is supposed to check the status through the response word 0 and interpret the other response data/ words accordingly.


slingappa:
Looks good to me

PR #91, opened 2026-05-26
lib: clock: return SUCCESS when the operation is a no-op

ch-perry:
First commit fixes a bug.

Second commit is a drive-by refactor.


pathakraul:
Looks good to me

PR #96, opened 2026-06-15
test: add CPPC service group coverage

slingappa:
This PR adds CPPC service-group test coverage.

Stack
- PR-1: https://github.com/riscv-software-src/librpmi/pull/86
- PR-2: https://github.com/riscv-software-src/librpmi/pull/92
- PR-3: https://github.com/riscv-software-src/librpmi/pull/93
- PR-4: https://github.com/riscv-software-src/librpmi/pull/94
- PR-5: https://github.com/riscv-software-src/librpmi/pull/95  ← base
- PR-6: https://github.com/riscv-software-src/librpmi/pull/96  ← this PR
- PR-7: https://github.com/riscv-software-src/librpmi/pull/97
- PR-8: https://github.com/riscv-software-src/librpmi/pull/98
- PR-9: https://github.com/riscv-software-src/librpmi/pull/99
- PR-10: https://github.com/riscv-software-src/librpmi/pull/100
- PR-11: https://github.com/riscv-software-src/librpmi/pull/101

Depends on: PR-5

Changes
- Add CPPC service-group tests for:
  - Notification unsupported behavior
  - Register probe/read/write paths
  - Fast-channel region and offset discovery
  - Hart-list discovery
  - Invalid register IDs, unsupported registers, invalid harts, and denied writes

Validation
- `make LIBRPMI_TEST=y CROSS_COMPILE= && make check CROSS_COMPILE=`

PR #94, opened 2026-06-15
test: add SYSTEM_MSI service group coverage

slingappa:
This PR adds SYSTEM_MSI service-group test coverage.

Stack
- PR-1: https://github.com/riscv-software-src/librpmi/pull/86
- PR-2: https://github.com/riscv-software-src/librpmi/pull/92
- PR-3: https://github.com/riscv-software-src/librpmi/pull/93  ← base
- PR-4: https://github.com/riscv-software-src/librpmi/pull/94  ← this PR
- PR-5: https://github.com/riscv-software-src/librpmi/pull/95
- PR-6: https://github.com/riscv-software-src/librpmi/pull/96
- PR-7: https://github.com/riscv-software-src/librpmi/pull/97
- PR-8: https://github.com/riscv-software-src/librpmi/pull/98
- PR-9: https://github.com/riscv-software-src/librpmi/pull/99
- PR-10: https://github.com/riscv-software-src/librpmi/pull/100
- PR-11: https://github.com/riscv-software-src/librpmi/pull/101

Depends on: PR-3

Changes
- Add SYSTEM_MSI service-group tests for:
  - Notification unsupported behavior
  - Group and MSI attributes
  - MSI state set/get
  - MSI target set/get
  - Invalid MSI index and invalid target address paths

Validation
- `make LIBRPMI_TEST=y CROSS_COMPILE= && make check CROSS_COMPILE=`

PR #93, opened 2026-06-15
test: expand HSM suspend service coverage

slingappa:
This PR expands HSM suspend service coverage.

Stack
- PR-1: https://github.com/riscv-software-src/librpmi/pull/86
- PR-2: https://github.com/riscv-software-src/librpmi/pull/92  ← base
- PR-3: https://github.com/riscv-software-src/librpmi/pull/93  ← this PR
- PR-4: https://github.com/riscv-software-src/librpmi/pull/94
- PR-5: https://github.com/riscv-software-src/librpmi/pull/95
- PR-6: https://github.com/riscv-software-src/librpmi/pull/96
- PR-7: https://github.com/riscv-software-src/librpmi/pull/97
- PR-8: https://github.com/riscv-software-src/librpmi/pull/98
- PR-9: https://github.com/riscv-software-src/librpmi/pull/99
- PR-10: https://github.com/riscv-software-src/librpmi/pull/100
- PR-11: https://github.com/riscv-software-src/librpmi/pull/101

Depends on: PR-2

Changes
- Extend existing HSM coverage for:
  - Suspend type listing when no suspend types are configured
  - Invalid suspend-info requests
  - Invalid and unsupported suspend requests

Validation
- `make LIBRPMI_TEST=y CROSS_COMPILE= && make check CROSS_COMPILE=`

PR #92, opened 2026-06-15
test: verify SYSTEM_RESET callback behavior

slingappa:
This PR adds focused SYSTEM_RESET callback behavior coverage.

Stack
- PR-1: https://github.com/riscv-software-src/librpmi/pull/86  ← base
- PR-2: https://github.com/riscv-software-src/librpmi/pull/92  ← this PR
- PR-3: https://github.com/riscv-software-src/librpmi/pull/93
- PR-4: https://github.com/riscv-software-src/librpmi/pull/94
- PR-5: https://github.com/riscv-software-src/librpmi/pull/95
- PR-6: https://github.com/riscv-software-src/librpmi/pull/96
- PR-7: https://github.com/riscv-software-src/librpmi/pull/97
- PR-8: https://github.com/riscv-software-src/librpmi/pull/98
- PR-9: https://github.com/riscv-software-src/librpmi/pull/99
- PR-10: https://github.com/riscv-software-src/librpmi/pull/100
- PR-11: https://github.com/riscv-software-src/librpmi/pull/101

Depends on: PR-1

Changes
- Extend existing SYSTEM_RESET coverage for:
  - Callback count/type validation
  - Cold reboot and shutdown reset paths
  - Unsupported reset type behavior
  - Posted-request no-ACK behavior

Validation
- `make LIBRPMI_TEST=y CROSS_COMPILE= && make check CROSS_COMPILE=`

PR #97, opened 2026-06-15
test: add Voltage service group coverage

slingappa:
This PR adds Voltage service-group test coverage.

Stack
- PR-1: https://github.com/riscv-software-src/librpmi/pull/86
- PR-2: https://github.com/riscv-software-src/librpmi/pull/92
- PR-3: https://github.com/riscv-software-src/librpmi/pull/93
- PR-4: https://github.com/riscv-software-src/librpmi/pull/94
- PR-5: https://github.com/riscv-software-src/librpmi/pull/95
- PR-6: https://github.com/riscv-software-src/librpmi/pull/96  ← base
- PR-7: https://github.com/riscv-software-src/librpmi/pull/97  ← this PR
- PR-8: https://github.com/riscv-software-src/librpmi/pull/98
- PR-9: https://github.com/riscv-software-src/librpmi/pull/99
- PR-10: https://github.com/riscv-software-src/librpmi/pull/100
- PR-11: https://github.com/riscv-software-src/librpmi/pull/101

Depends on: PR-6

Changes
- Add Voltage service-group tests for:
  - Domain count and attributes
  - Discrete and linear supported-level queries
  - Config set/get
  - Voltage level set/get
  - Invalid domains, short requests, bad config, and unsupported levels

Validation
- `make LIBRPMI_TEST=y CROSS_COMPILE= && make check CROSS_COMPILE=`

### DIFF
--- a/test/test_srvgrp_voltage.c
+++ b/test/test_srvgrp_voltage.c
@@ -1,0 +1,595 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2024 Ventana Micro Systems Inc.
+ */
+
+#include <librpmi.h>
+#include <stdio.h>
+#include "test_common.h"
+#include "test_log.h"
+
+#define TEST_VOLT_COUNT		2
+#define TEST_VOLT_CPU		0
+#define TEST_VOLT_MEM		1
+#define TEST_VOLT_INVALID	TEST_VOLT_COUNT
+#define TEST_EVENT_ID		0x0
+#define TEST_REQUEST_STATE_ENABLE	0x1
+#define TEST_VOLT_CPU_LATENCY	10
+#define TEST_VOLT_MEM_LATENCY	20
+#define TEST_VOLT_CPU_INITIAL	800000
+#define TEST_VOLT_CPU_TARGET	1000000
+#define TEST_VOLT_CPU_UNSUPP	1100000
+#define TEST_VOLT_BAD_CONFIG	RPMI_VOLT_CONFIG_MAX
+
+struct test_volt_attrs_expdata {
+	rpmi_uint32_t status;
+	rpmi_uint32_t capability;
+	rpmi_uint32_t num_levels;
+	rpmi_uint32_t trans_latency;
+	char name[16];
+};
+
+static rpmi_int32_t test_cpu_levels[] = {
+	800000,
+	900000,
+	1000000,
+};
+
+static rpmi_int32_t test_mem_levels[] = {
+	1800000,
+	1900000,
+};
+
+static struct rpmi_voltage_discrete_range test_cpu_range = {
+	.uvolt = (rpmi_uint32_t *)test_cpu_levels,
+};
+
+static struct rpmi_voltage_linear_range test_mem_range = {
+	.uvolt_min = 1800000,
+	.uvolt_max = 1900000,
+	.uvolt_step = 100000,
+};
+
+static struct rpmi_voltage_data test_voltage_data[TEST_VOLT_COUNT] = {
+	[TEST_VOLT_CPU] = {
+		.name = "vdd_cpu",
+		.voltage_type = RPMI_VOLT_TYPE_DISCRETE,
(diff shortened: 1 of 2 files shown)


pathakraul:
@slingappa I see commits in this PR which should belong to other PRs and some of then would have merged. How are we managing this series of dependent PRs

slingappa:
@pathakraul please review

test/test_srvgrp_voltage.c
```diff
@@ -0,0 +1,562 @@
...
+				.request_data = volt_cpu_reqdata,
+				.request_data_len = sizeof(volt_cpu_reqdata),
+				.expected_data = get_config_enabled_expdata,
+				.expected_data_len = sizeof(get_config_enabled_expdata),
+			},
+			.init_request_data = test_init_request_data_from_attrs,
+			.init_expected_data = test_init_expected_data_from_attrs,
+		},
+		{
+			.name = "GET CONFIG (short request)",
```
pathakraul: What is "short request"

test/test_srvgrp_voltage.c
```diff
@@ -0,0 +1,562 @@
...
+	TEST_VOLT_CPU,
+};
+
+static rpmi_uint32_t volt_invalid_reqdata[] = {
+	TEST_VOLT_INVALID,
+};
+
+static struct test_volt_attrs_expdata get_attrs_cpu_expdata = {
+	.status = RPMI_SUCCESS,
+	.capability = RPMI_VOLT_TYPE_DISCRETE | RPMI_VOLT_CAPABILITY_ENABLED_DISABLED,
```
pathakraul: The Voltage format and the Voltage control occupy different bit fields in the capability word. And currently this does not shift the voltage format to its actual bit location.

I checked and same is done in the service group implementation which is a bug and needs to be patched

test/test_srvgrp_voltage.c
slingappa: Updated the test name to `GET CONFIG (missing domain id)` so it describes the case more directly. The request intentionally omits the domain ID, which makes the payload shorter than the service's minimum request length and should return `RPMI_ERR_NOTSUPP`.

test/test_srvgrp_voltage.c
slingappa: Split the service implementation fix into PR #103. `RPMI_VOLT_TYPE_LINEAR` is now the unshifted voltage format value there, and the capability word is built with helpers that place the format in bits `[3:1]` and control in bit `[0]`. PR 97 is now based on that lib-fix branch and keeps the Voltage test coverage update, including `GET ATTRIBUTES (linear)` coverage for the non-zero encoded capability value.

slingappa:
@pathakraul Re: stacked PR management question in https://github.com/riscv-software-src/librpmi/pull/97#issuecomment-4769614199

This PR is part of the stacked test series and is now based on `stack/slingapp_tests2/pr-7-voltage-capability-fix` / PR #103. As lower PRs merge, I am rebasing each dependent branch onto the updated lower branch/main using `git rebase --onto <new-base> <old-parent> <branch>` and then updating the PR branch with `--force-with-lease`. PR 97 should only carry the Voltage service group test delta over the split-out lib fix.

Addressed the review comments by clarifying the missing-domain-ID test name and splitting the voltage capability encoding fix into PR #103. PR 97 now adds linear `GET ATTRIBUTES` coverage to exercise the non-zero encoded capability value.

test/test_srvgrp_voltage.c
pathakraul: What do you mean by missing domain id?

Even if you are not writing a valid domain id in the message, the domain id field will contain a garbage number which will be considered an invalid domain id (not missing domain id). 

The missing/invalid domain id should return RPMI_ERR_INVALID_PARAM not RPMI_ERR_NOTSUPP.

slingappa: Good point, thanks. I updated this case to send an explicit invalid domain ID instead of omitting the request payload. The test is now named `GET CONFIG (invalid domain)`, initializes `volt_invalid_reqdata`, and expects `RPMI_ERR_INVALID_PARAM`.

PR #86, opened 2026-05-15
test: improve RPMI harness and expand BASE coverage

slingappa:
This PR keeps the original PR #86 history while narrowing it to the RPMI test harness and BASE service-group coverage changes.

Stack
- PR-1: https://github.com/riscv-software-src/librpmi/pull/86  ← this PR
- PR-2: https://github.com/riscv-software-src/librpmi/pull/92
- PR-3: https://github.com/riscv-software-src/librpmi/pull/93
- PR-4: https://github.com/riscv-software-src/librpmi/pull/94
- PR-5: https://github.com/riscv-software-src/librpmi/pull/95
- PR-6: https://github.com/riscv-software-src/librpmi/pull/96
- PR-7: https://github.com/riscv-software-src/librpmi/pull/97
- PR-8: https://github.com/riscv-software-src/librpmi/pull/98
- PR-9: https://github.com/riscv-software-src/librpmi/pull/99
- PR-10: https://github.com/riscv-software-src/librpmi/pull/100
- PR-11: https://github.com/riscv-software-src/librpmi/pull/101

Changes
- Update the shared RPMI test harness so failed comparisons cause test ELF failure and make check failure.
- Extend existing BASE coverage for:
  - ACK header validation
  - Unsupported service-group probes
  - Experimental/vendor service-group probe cases
  - Invalid BASE service ID handling

Validation
- `make LIBRPMI_TEST=y CROSS_COMPILE= && make check CROSS_COMPILE=`

PR #98, opened 2026-06-15
test: add Clock service group coverage

slingappa:
This PR adds Clock service-group test coverage.

Stack
- PR-1: https://github.com/riscv-software-src/librpmi/pull/86
- PR-2: https://github.com/riscv-software-src/librpmi/pull/92
- PR-3: https://github.com/riscv-software-src/librpmi/pull/93
- PR-4: https://github.com/riscv-software-src/librpmi/pull/94
- PR-5: https://github.com/riscv-software-src/librpmi/pull/95
- PR-6: https://github.com/riscv-software-src/librpmi/pull/96
- PR-7: https://github.com/riscv-software-src/librpmi/pull/97  ← base
- PR-8: https://github.com/riscv-software-src/librpmi/pull/98  ← this PR
- PR-9: https://github.com/riscv-software-src/librpmi/pull/99
- PR-10: https://github.com/riscv-software-src/librpmi/pull/100
- PR-11: https://github.com/riscv-software-src/librpmi/pull/101

Depends on: PR-7

Changes
- Add Clock service-group tests for:
  - Clock count and attributes
  - Discrete and linear supported-rate queries
  - Config set/get
  - 64-bit rate set/get
  - Invalid clocks, bad indexes, bad rate-match flags, zero rates, and unsupported rates

Validation
- `make LIBRPMI_TEST=y CROSS_COMPILE= && make check CROSS_COMPILE=`

### DIFF
--- a/test/test_srvgrp_clock.c
+++ b/test/test_srvgrp_clock.c
@@ -1,0 +1,581 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2024 Ventana Micro Systems Inc.
+ */
+
+#include <librpmi.h>
+#include <stdio.h>
+#include "test_common.h"
+#include "test_log.h"
+
+#define TEST_CLK_COUNT		2
+#define TEST_CLK_CPU		0
+#define TEST_CLK_BUS		1
+#define TEST_CLK_INVALID	TEST_CLK_COUNT
+#define TEST_EVENT_ID		0x0
+#define TEST_REQUEST_STATE_ENABLE	0x1
+#define TEST_CLK_CPU_LATENCY	5
+#define TEST_CLK_BUS_LATENCY	7
+#define TEST_CLK_CPU_INITIAL_RATE	100000000ULL
+#define TEST_CLK_CPU_TARGET_RATE	5000000000ULL
+#define TEST_CLK_CPU_MAX_RATE		6000000000ULL
+#define TEST_CLK_CPU_UNSUPP_RATE	7000000000ULL
+#define TEST_CLK_BUS_MIN_RATE	500000000ULL
+#define TEST_CLK_BUS_MAX_RATE	1000000000ULL
+#define TEST_CLK_BUS_STEP_RATE	100000000ULL
+
+#define RATE_LO(_rate)	((rpmi_uint32_t)((rpmi_uint64_t)(_rate)))
+#define RATE_HI(_rate)	((rpmi_uint32_t)(((rpmi_uint64_t)(_rate)) >> 32))
+
+struct test_clk_attrs_expdata {
+	rpmi_uint32_t status;
+	rpmi_uint32_t flags;
+	rpmi_uint32_t rate_count;
+	rpmi_uint32_t transition_latency;
+	char name[16];
+};
+
+static const rpmi_uint64_t test_cpu_rates[] = {
+	TEST_CLK_CPU_INITIAL_RATE,
+	TEST_CLK_CPU_TARGET_RATE,
+	TEST_CLK_CPU_MAX_RATE,
+};
+
+static const rpmi_uint64_t test_bus_rates[] = {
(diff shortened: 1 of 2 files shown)


test/test_srvgrp_clock.c
```diff
@@ -0,0 +1,573 @@
...
+	rpmi_uint32_t flags;
+	rpmi_uint32_t rate_count;
+	rpmi_uint32_t transition_latency;
+	char name[16];
+};
+
+static const rpmi_uint64_t test_cpu_rates[] = {
+	TEST_CLK_CPU_INITIAL_RATE,
+	TEST_CLK_CPU_TARGET_RATE,
+	400000000ULL,
```
pathakraul: NIT: better to keep define as same, macro or scaler value

pathakraul: Same at other places

test/test_srvgrp_clock.c
```diff
@@ -0,0 +1,573 @@
...
+#define TEST_EVENT_ID		0x0
+#define TEST_REQUEST_STATE_ENABLE	0x1
+#define TEST_CLK_CPU_LATENCY	5
+#define TEST_CLK_BUS_LATENCY	7
+#define TEST_CLK_CPU_INITIAL_RATE	100000000ULL
+#define TEST_CLK_CPU_TARGET_RATE	200000000ULL
+#define TEST_CLK_CPU_UNSUPP_RATE	123456789ULL
+#define TEST_CLK_BUS_MIN_RATE	500000000ULL
+#define TEST_CLK_BUS_MAX_RATE	1000000000ULL
+#define TEST_CLK_BUS_STEP_RATE	100000000ULL
```
pathakraul, line 25: All rates are well below 2^32, may be for some variation, any rate can be defined which is greater than UNIT32_MAX

slingappa: Addressed in b1ce454. I moved the exercised CPU target/max and unsupported rates above `UINT32_MAX`, so the test now covers non-zero `RATE_HI()` values in the supported-rate path and in the set/get-rate requests. The updated clock test still passes in the host out-of-tree build.

test/test_srvgrp_clock.c
slingappa: Addressed in b1ce454. I replaced the raw `400000000ULL` literal with `TEST_CLK_CPU_MAX_RATE` and used the same macro in both `test_cpu_rates[]` and `get_cpu_rates_expdata[]`, so the discrete-rate data stays consistent across the test. This thread is now outdated after the squashed force-push.

test/test_srvgrp_clock.c
```diff
@@ -0,0 +1,574 @@
...
+				.request_data = clk_cpu_reqdata,
+				.request_data_len = sizeof(clk_cpu_reqdata),
+				.expected_data = get_config_enabled_expdata,
+				.expected_data_len = sizeof(get_config_enabled_expdata),
+			},
+			.init_request_data = test_init_request_data_from_attrs,
+			.init_expected_data = test_init_expected_data_from_attrs,
+		},
+		{
+			.name = "GET CONFIG (short request)",
```
pathakraul: According to this test, a short request means that Request Message Datalen == 0 even though there is a request message data in the payload?

pathakraul: What if the test harness or the underlying RPMI core skips sending the message since `.request_data` is not initialized along with `.init_request_data`

slingappa: Applied in 6c07d5c. Updated this test so it no longer relies on an omitted request body. It now sends a one-byte GET CONFIG payload through test_init_request_data_from_attrs, so the harness enqueues a normal request with a non-zero datalen that is still shorter than the 4-byte clock-id minimum. I also renamed it to GET CONFIG (truncated clock id) to make the intent explicit. The clock scenario passes with this change.

pathakraul:
Looks good to me

PR #102, opened 2026-06-23
fix typos in the device power service group

ch-perry:


pathakraul:
Typos could have been in a single commit instead of two since its the same file.

Looks good to me

ch-perry:
Ok, thanks for the feedback!

PR #104, opened 2026-07-02
lib: hsm: process deferred state before transitions

slingappa:
## Summary
- process deferred HSM hart state changes before validating start, stop, and suspend requests
- drain a pending stop briefly before rejecting a subsequent hart start request
- mirrors the fix validated in QEMU for intermittent kvm-unit-tests SBI HSM hangs on the RPMI HSM path

## Validation
- make LIBRPMI_TEST=y
- env -u CROSS_COMPILE make O=build-native LIBRPMI_TEST=y CC=gcc AR=ar AS=as LD=ld
- env -u CROSS_COMPILE make O=build-native check CC=gcc AR=ar AS=as LD=ld

Note: plain make check in this environment used the default RISC-V cross compiler and produced Exec format error on the x86 host, so native check was run explicitly with gcc.

### DIFF
--- a/lib/rpmi_hsm.c
+++ b/lib/rpmi_hsm.c
@@ -12,6 +12,13 @@
 #define DPRINTF(msg...)
 #endif
 
+/*
+ * Bound the polling loop used to refresh a pending STOP state before START.
+ * This is not a time value; each iteration samples the platform HW-state hook
+ * once while holding the hart lock.
+ */
+#define RPMI_HSM_STOP_PENDING_RETRIES	10000
+
 struct rpmi_hsm_hart {
 	/** Lock to protect this structure and perform platform operations */
 	void *lock;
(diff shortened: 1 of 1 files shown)


lib/rpmi_hsm.c
```diff
@@ -288,6 +288,16 @@ enum rpmi_error rpmi_hsm_hart_start(struct rpmi_hsm *hsm, rpmi_uint32_t hart_id,
 
 	hart = &hsm->leaf.harts[hart_index];
 	rpmi_env_lock(hart->lock);
+	__rpmi_hsm_process_hart_state_changes(hsm, hart, hart_index);
+	if (hart->state == RPMI_HSM_HART_STATE_STOP_PENDING) {
+		rpmi_uint32_t retries;
+
+		for (retries = 0; retries < 10000 &&
```
pathakraul: What is the significance of `10000`

slingappa: Addressed in `8a8e0fc`. I replaced the raw `10000` with `RPMI_HSM_STOP_PENDING_RETRIES` and documented that it is a bounded poll count, not a time value. Each retry samples the platform HW-state hook once while waiting for `STOP_PENDING` to clear before starting the hart.

pathakraul:
Looks good

PR #105, opened 2026-07-03
lib: base: implement ENABLE_NOTIFICATION service

icosta-sifive:
This PR implements the ENABLE_NOTIFICATION service for the BASE service group as per the RPMI specification.

## Overview

Adds support for event notifications in the BASE service group, allowing the application processor to subscribe to REQUEST_HANDLE_ERROR events. This is useful for detecting when the platform firmware cannot handle an AP request (e.g., when P2A ACK enqueue fails).

### Library Implementation (`lib/rpmi_context.c`)
- Implemented `rpmi_base_enable_notification()` handler to enable/disable notifications based on EVENT_ID and REQ_STATE parameters
- Added `rpmi_base_process_events()` callback to send P2A notification messages when events are pending
- Extended `struct rpmi_base_group` with notification state tracking (enabled flag, pending event flag, pre-allocated message buffer)
- Updated `rpmi_base_get_attributes()` to set `RPMI_BASE_FLAGS_F0_EV_NOTIFY` flag
- Modified `rpmi_context_process_a2p_request()` to trigger REQUEST_HANDLE_ERROR when P2A ACK enqueue fails

### Public API (`include/librpmi.h`)
- Added `enum rpmi_base_event_id` with `RPMI_BASE_EVENT_REQUEST_HANDLE_ERROR`
- Added `rpmi_context_base_request_handle_error()` API for platform firmware to trigger notification events

### Tests (`test/test_srvgrp_base.c`)
- Updated ENABLE_NOTIFICATION test to verify successful enablement
- Updated GET_ATTRIBUTES test to expect EV_NOTIFY flag

## Testing

All existing tests continue to pass. New notification functionality is verified through updated base service group tests.

## Compliance

Implements BASE service group ENABLE_NOTIFICATION service as specified in the RISC-V RPMI specification.

### DIFF
--- a/lib/rpmi_context.c
+++ b/lib/rpmi_context.c
@@ -58,6 +58,15 @@ struct rpmi_base_group {
 	/* Platform info string */
 	char *plat_info;
 
+	/* Notification enabled state for REQUEST_HANDLE_ERROR event */
+	rpmi_bool_t notif_enabled;
+
+	/* Pending REQUEST_HANDLE_ERROR notification flag */
+	rpmi_bool_t pending_request_handle_err;
+
+	/* Pre-allocated notification message buffer */
+	struct rpmi_message *notif_msg;
+
 	struct rpmi_service_group group;
 };
 
@@ -163,6 +172,60 @@ static enum rpmi_error rpmi_base_probe_group(struct rpmi_service_group *group,
 	return RPMI_SUCCESS;
 }
 
+static enum rpmi_error rpmi_base_enable_notification(struct rpmi_service_group *group,
+						     struct rpmi_service *service,
+						     struct rpmi_transport *trans,
+						     rpmi_uint16_t request_datalen,
+						     const rpmi_uint8_t *request_data,
+						     rpmi_uint16_t *response_datalen,
+						     rpmi_uint8_t *response_data)
+{
+	const rpmi_uint32_t *req = (const rpmi_uint32_t *)request_data;
+	rpmi_uint32_t *resp = (void *)response_data;
+	struct rpmi_base_group *base = group->priv;
+	rpmi_uint32_t event_id, req_state;
+
+	/* Notifications require a P2A channel to deliver events */
+	if (!trans->is_p2a_channel) {
+		*response_datalen = sizeof(*resp);
+		resp[0] = rpmi_to_xe32(trans->is_be, (rpmi_uint32_t)RPMI_ERR_NOTSUPP);
+		return RPMI_SUCCESS;
+	}
+
+	event_id = rpmi_to_xe32(trans->is_be, req[0]);
+	req_state = rpmi_to_xe32(trans->is_be, req[1]);
+
+	/* Only REQUEST_HANDLE_ERROR event is supported */
+	if (event_id != RPMI_BASE_EVENT_REQUEST_HANDLE_ERROR) {
+		*response_datalen = sizeof(*resp);
+		resp[0] = rpmi_to_xe32(trans->is_be, (rpmi_uint32_t)RPMI_ERR_INVALID_PARAM);
+		return RPMI_SUCCESS;
+	}
+
+	switch (req_state) {
+	case 0: /* Disable */
+		base->notif_enabled = false;
+		base->pending_request_handle_err = false;
+		break;
+	case 1: /* Enable */
+		base->notif_enabled = true;
+		break;
+	case 2: /* Query current state */
+		break;
+	default:
+		*response_datalen = sizeof(*resp);
+		resp[0] = rpmi_to_xe32(trans->is_be, (rpmi_uint32_t)RPMI_ERR_INVALID_PARAM);
+		return RPMI_SUCCESS;
+	}
+
+	/* Return current state in response */
+	*response_datalen = 2 * sizeof(*resp);
+	resp[0] = rpmi_to_xe32(trans->is_be, (rpmi_uint32_t)RPMI_SUCCESS);
+	resp[1] = rpmi_to_xe32(trans->is_be, base->notif_enabled ? 1U : 0U);
+
+	return RPMI_SUCCESS;
+}
+
 static enum rpmi_error rpmi_base_get_attributes(struct rpmi_service_group *group,
 						struct rpmi_service *service,
 						struct rpmi_transport *trans,
@@ -179,6 +242,10 @@ static enum rpmi_error rpmi_base_get_attributes(struct rpmi_service_group *group
 	flags |= (cntx->privilege_level == RPMI_PRIVILEGE_M_MODE) ?
 					RPMI_BASE_FLAGS_F0_PRIVILEGE : 0;
 
+	/* Set the event notification support flag only if transport has P2A channel */
+	if (trans->is_p2a_channel)
+		flags |= RPMI_BASE_FLAGS_F0_EV_NOTIFY;
+
 	*response_datalen = 5 * sizeof(*resp);
 	resp[0] = rpmi_to_xe32(trans->is_be, (rpmi_uint32_t)RPMI_SUCCESS);
 	resp[1] = rpmi_to_xe32(trans->is_be, flags);
@@ -189,11 +256,51 @@ static enum rpmi_error rpmi_base_get_attributes(struct rpmi_service_group *group
 	return RPMI_SUCCESS;
 }
 
+static enum rpmi_error rpmi_base_process_events(struct rpmi_service_group *group)
+{
+	struct rpmi_base_group *base = group->priv;
+	struct rpmi_transport *trans = base->cntx->trans;
+	struct rpmi_message *notif_msg = base->notif_msg;
+	rpmi_uint32_t *data;
+	enum rpmi_error rc;
+
+	/* Check if notification is enabled and there's a pending event */
+	if (!base->notif_enabled || !base->pending_request_handle_err)
+		return RPMI_SUCCESS;
+
+	/* Prepare P2A notification message */
+	notif_msg->header.servicegroup_id = RPMI_SRVGRP_BASE;
+	notif_msg->header.service_id = 0; /* Not used for notifications */
+	notif_msg->header.flags = RPMI_MSG_NOTIFICATION;
+	notif_msg->header.token = 0; /* Not used for notifications */
+	notif_msg->header.datalen = 2 * sizeof(rpmi_uint32_t);
+
+	data = (rpmi_uint32_t *)notif_msg->data;
+	data[0] = rpmi_to_xe32(trans->is_be, RPMI_BASE_EVENT_REQUEST_HANDLE_ERROR);
+	data[1] = rpmi_to_xe32(trans->is_be, 0); /* Event-specific data (reserved) */
+
+	/* Try to enqueue the notification */
+	rc = rpmi_transport_enqueue(trans, RPMI_QUEUE_P2A_REQ, notif_msg);
+	if (rc == RPMI_SUCCESS) {
+		/* Clear the pending flag on successful enqueue */
+		base->pending_request_handle_err = false;
+		return RPMI_SUCCESS;
+	} else if (rc == RPMI_ERR_IO) {
+		/* Queue is full, will retry later */
+		return RPMI_ERR_BUSY;
+	}
+
+	/* Other error occurred */
+	DPRINTF("%s: failed to enqueue notification (error %d)\n",
+		__func__, rc);
+	return rc;
+}
+
 static struct rpmi_service rpmi_base_services[RPMI_BASE_SRV_ID_MAX] = {
 	[RPMI_BASE_SRV_ENABLE_NOTIFICATION] = {
 		.service_id = RPMI_BASE_SRV_ENABLE_NOTIFICATION,
 		.min_a2p_request_datalen = 8,
-		.process_a2p_request = NULL,
+		.process_a2p_request = rpmi_base_enable_notification,
 	},
 	[RPMI_BASE_SRV_GET_IMPLEMENTATION_VERSION] = {
 		.service_id = RPMI_BASE_SRV_GET_IMPLEMENTATION_VERSION,
@@ -262,6 +369,15 @@ static struct rpmi_service_group *rpmi_base_group_create(struct rpmi_context *cn
 		rpmi_env_strncpy(base->plat_info, plat_info, plat_info_len);
 	}
 
+	/* Initialize notification state */
+	base->notif_enabled = false;
+	base->pending_request_handle_err = false;
+
+	/* Allocate pre-allocated notification message buffer */
+	base->notif_msg = rpmi_env_zalloc(cntx->trans->slot_size);
+	if (!base->notif_msg)
+		goto fail_free_plat_info;
+
 	group = &base->group;
 	group->name = "base";
 	group->servicegroup_id = RPMI_SRVGRP_BASE;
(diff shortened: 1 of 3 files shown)


icosta-sifive:
@pathakraul @avpatel Same on this one, this PR implements BASE service group notifications - a missing feature from the RPMI spec.

Adds ENABLE_NOTIFICATION support for REQUEST_HANDLE_ERROR events, including P2A messaging and error handling. All tests pass.

I don't have reviewer permissions - could you help get this reviewed? Thanks! 🙏

pathakraul:
@icosta-sifive Will be reviewed without adding the reviewers

pathakraul:
How are you testing this? Do you have an counterpart Linux support for testing

icosta-sifive:
> How are you testing this? Do you have an counterpart Linux support for testing

@pathakraul I do not have a Linux counterpart for this for the time being. So our our internal tools we have validated end-to-end by running a bare-metal AP stub on a RISC-V QEMU model where platform firmware and AP run as separate QEMU instances communicating over real shared-memory mailbox registers and doorbell interrupts. 
The following 9 steps are exercised in sequence:

| Step | Request | What is verified |
|-----:|---------|-----------------|
| 0 | `GET_ATTRIBUTES` | `F0_EV_NOTIFY` flag is advertised |
| 1 | `ENABLE_NOTIFICATION(event=REQUEST_HANDLE_ERROR, state=enable)` | `status=0`, `enabled_state=1` |
| 2 | `ENABLE_NOTIFICATION(event=REQUEST_HANDLE_ERROR, state=query)` | `enabled_state=1` (unchanged) |
| 3 | `ENABLE_NOTIFICATION(event=0xDEAD, state=enable)` | `status=RPMI_ERR_INVALID_PARAM` |
| 4 | 72 × `GET_IMPLEMENTATION_VERSION` (no ACK drain) | P2A\_ACK queue + buffer exhausted without error |
| 5 | Poll `P2A_REQ` | `REQUEST_HANDLE_ERROR` notification received, `event_id=0x01` |
| 6 | Drain stale ACKs | `P2A_ACK` cleared for next requests |
| 7 | `ENABLE_NOTIFICATION(event=REQUEST_HANDLE_ERROR, state=disable)` | `status=0`, `enabled_state=0` |
| 8 | `ENABLE_NOTIFICATION(event=REQUEST_HANDLE_ERROR, state=query)` | `enabled_state=0` |
| 9 | 72 × `GET_IMPLEMENTATION_VERSION` (no ACK drain) | No notification received (disable effective) |

pathakraul:
> > How are you testing this? Do you have an counterpart Linux support for testing
> 
> @pathakraul I do not have a Linux counterpart for this for the time being. So our our internal tools we have validated end-to-end by running a bare-metal AP stub on a RISC-V QEMU model where platform firmware and AP run as separate QEMU instances communicating 
over real shared-memory mailbox registers and doorbell interrupts. The following 9 steps are exercised in sequence:

Do you have the baremetal Qemu stubs posted somewhere.


> 
> Step	Request	What is verified
> 0	`GET_ATTRIBUTES`	`F0_EV_NOTIFY` flag is advertised
> 1	`ENABLE_NOTIFICATION(event=REQUEST_HANDLE_ERROR, state=enable)`	`status=0`, `enabled_state=1`
> 2	`ENABLE_NOTIFICATION(event=REQUEST_HANDLE_ERROR, state=query)`	`enabled_state=1` (unchanged)
> 3	`ENABLE_NOTIFICATION(event=0xDEAD, state=enable)`	`status=RPMI_ERR_INVALID_PARAM`
> 4	72 × `GET_IMPLEMENTATION_VERSION` (no ACK drain)	P2A_ACK queue + buffer exhausted without error
> 5	Poll `P2A_REQ`	`REQUEST_HANDLE_ERROR` notification received, `event_id=0x01`
> 6	Drain stale ACKs	`P2A_ACK` cleared for next requests
> 7	`ENABLE_NOTIFICATION(event=REQUEST_HANDLE_ERROR, state=disable)`	`status=0`, `enabled_state=0`
> 8	`ENABLE_NOTIFICATION(event=REQUEST_HANDLE_ERROR, state=query)`	`enabled_state=0`
> 9	72 × `GET_IMPLEMENTATION_VERSION` (no ACK drain)	No notification received (disable effective)

icosta-sifive:
> > > How are you testing this? Do you have an counterpart Linux support for testing
> > 
> > 
> > @pathakraul I do not have a Linux counterpart for this for the time being. So our our internal tools we have validated end-to-end by running a bare-metal AP stub on a RISC-V QEMU model where platform firmware and AP run as separate QEMU instances communicating
> > over real shared-memory mailbox registers and doorbell interrupts. The following 9 steps are exercised in sequence:
> 
> Do you have the baremetal Qemu stubs posted somewhere.
> 
> > Step	Request	What is verified
> > 0	`GET_ATTRIBUTES`	`F0_EV_NOTIFY` flag is advertised
> > 1	`ENABLE_NOTIFICATION(event=REQUEST_HANDLE_ERROR, state=enable)`	`status=0`, `enabled_state=1`
> > 2	`ENABLE_NOTIFICATION(event=REQUEST_HANDLE_ERROR, state=query)`	`enabled_state=1` (unchanged)
> > 3	`ENABLE_NOTIFICATION(event=0xDEAD, state=enable)`	`status=RPMI_ERR_INVALID_PARAM`
> > 4	72 × `GET_IMPLEMENTATION_VERSION` (no ACK drain)	P2A_ACK queue + buffer exhausted without error
> > 5	Poll `P2A_REQ`	`REQUEST_HANDLE_ERROR` notification received, `event_id=0x01`
> > 6	Drain stale ACKs	`P2A_ACK` cleared for next requests
> > 7	`ENABLE_NOTIFICATION(event=REQUEST_HANDLE_ERROR, state=disable)`	`status=0`, `enabled_state=0`
> > 8	`ENABLE_NOTIFICATION(event=REQUEST_HANDLE_ERROR, state=query)`	`enabled_state=0`
> > 9	72 × `GET_IMPLEMENTATION_VERSION` (no ACK drain)	No notification received (disable effective)

@pathakraul As mention previously, this is all internal tools/code repositories, so unfortunately i cannot disclosure them . Has we only have for now unit tests under librpmi, is difficult to upstream more than what i purpose on this PR. Also probably will be difficult to ever have a System tests on this repo.

lib/rpmi_context.c
```diff
@@ -163,6 +172,52 @@ static enum rpmi_error rpmi_base_probe_group(struct rpmi_service_group *group,
...
+	/* Only REQUEST_HANDLE_ERROR event is supported */
+	if (event_id != RPMI_BASE_EVENT_REQUEST_HANDLE_ERROR) {
+		*response_datalen = sizeof(*resp);
+		resp[0] = rpmi_to_xe32(trans->is_be, (rpmi_uint32_t)RPMI_ERR_INVALID_PARAM);
+		return RPMI_SUCCESS;
+	}
+
+	switch (req_state) {
+	case 0: /* Disable */
+		base->notif_enabled = false;
```
slingappa, line 207: When handling `REQ_STATE=0`, should this also clear `pending_request_handle_err`? I reproduced this through a real QEMU RPMI path using a P2A-enabled QEMU build: enable notification, verify `REQUEST_HANDLE_ERROR` delivery works, disable notification, trigger an ACK enqueue failure, drain ACKs, then re-enable notification. The old `REQUEST_HANDLE_ERROR` is delivered immediately after re-enable because `rpmi_base_process_events()` only clears the pending flag after a successful enqueue.

That means an event latched while notification was disabled can be reported after a later enable. If this delayed delivery is intentional per the spec, please document/test that semantic explicitly; otherwise please clear `pending_request_handle_err` when disabling the event.

lib/rpmi_context.c
```diff
@@ -179,6 +234,9 @@ static enum rpmi_error rpmi_base_get_attributes(struct rpmi_service_group *group
 	flags |= (cntx->privilege_level == RPMI_PRIVILEGE_M_MODE) ?
 					RPMI_BASE_FLAGS_F0_PRIVILEGE : 0;
 
+	/* Set the event notification support flag */
+	flags |= RPMI_BASE_FLAGS_F0_EV_NOTIFY;
```
slingappa: `GET_ATTRIBUTES` now advertises `RPMI_BASE_FLAGS_F0_EV_NOTIFY` unconditionally, and `ENABLE_NOTIFICATION` can succeed even when the transport has no P2A request channel. In that configuration, `rpmi_transport_enqueue(..., RPMI_QUEUE_P2A_REQ, ...)` rejects notification delivery because `trans->is_p2a_channel` is false, so the AP can enable an event that can never be delivered.

I hit this with the existing QEMU virt RPMI configuration before enabling the P2A request queue: the DT exposed only `a2p-req`, `p2a-ack`, and `a2p-doorbell`; `ENABLE_NOTIFICATION` succeeded, but an enabled `REQUEST_HANDLE_ERROR` notification never reached the AP.

Can notification support be advertised/enabled only when the transport has a P2A channel, or should this case return an error from `ENABLE_NOTIFICATION`?

test/test_srvgrp_base.c
```diff
@@ -22,10 +22,105 @@ static struct plat_info {
...
+static void test_base_notif_wait(struct rpmi_test_scenario *scene,
+				 struct rpmi_test *test,
+				 struct rpmi_message *msg)
+{
+	int result = -1;
+
+	rpmi_env_memset(msg, 0, scene->slot_size);
+
+	/* Notification arrives on P2A_REQ queue, not P2A_ACK */
+	while (result != RPMI_SUCCESS)
```
slingappa: This wait loop can spin forever if the notification path regresses and no message reaches `RPMI_QUEUE_P2A_REQ`. Could you add a bounded retry/failure path so the test fails with a useful diagnostic instead of hanging the test process?

lib/rpmi_context.c
```diff
@@ -189,11 +247,52 @@ static enum rpmi_error rpmi_base_get_attributes(struct rpmi_service_group *group
...
+
+	/* Try to enqueue the notification */
+	rc = rpmi_transport_enqueue(trans, RPMI_QUEUE_P2A_REQ, notif_msg);
+	if (rc == RPMI_SUCCESS) {
+		/* Clear the pending flag on successful enqueue */
+		base->pending_request_handle_err = false;
+	} else if (rc == RPMI_ERR_IO) {
+		/* Queue is full, will retry later */
+		return RPMI_ERR_BUSY;
+	} else {
```
slingappa: Optional nit from `checkpatch.pl`: the `else` after `return RPMI_ERR_BUSY;` is unnecessary. Please drop the `else` and keep the remaining error path after the early return.

slingappa (commented):
.

lib/rpmi_context.c
icosta-sifive: @slingappa Good catch. Fixed in the latest commit: case 0 (disable) now clears pending_request_handle_err alongside notif_enabled, so any latched-but-undelivered error is dropped when the AP opts out of notifications.

Added a regression test in scenario_base_notif_clear_on_disable that exercises the exact sequence you described: enable → latch pending_request_handle_err while notifications are enabled → disable → re-enable → process_all_events. The test asserts P2A_REQ is empty afterwards; it fails on the unfixed code and passes with the fix.

lib/rpmi_context.c
icosta-sifive: GET_ATTRIBUTES ([lib/rpmi_context.c](vscode-webview://0qamqpql5l1qndroi9r2vpg9pe6q5a8r709dhk101rug1ilrv09s/lib/rpmi_context.c)): RPMI_BASE_FLAGS_F0_EV_NOTIFY is now only set when trans->is_p2a_channel is true, so the AP gets an honest capability advertisement upfront.

ENABLE_NOTIFICATION ([lib/rpmi_context.c](vscode-webview://0qamqpql5l1qndroi9r2vpg9pe6q5a8r709dhk101rug1ilrv09s/lib/rpmi_context.c)): Added a guard at the top of the handler that returns RPMI_ERR_NOTSUPP immediately when !trans->is_p2a_channel, covering APs that call it without checking attributes first.

Added scenario_base_no_p2a_channel in the test suite (transport created with p2a_req_queue_size=0) that verifies both: GET_ATTRIBUTES returns flags without EV_NOTIFY, and ENABLE_NOTIFICATION returns RPMI_ERR_NOTSUPP.

test/test_srvgrp_base.c
icosta-sifive: Fixed. test_base_notif_wait now uses a bounded retry loop (up to TEST_DEQUEUE_MAX_RETRIES = 1000 attempts) and prints a clear diagnostic if the limit is reached before a message arrives:


<test name>: timed out waiting for notification on RPMI_QUEUE_P2A_REQ after 1000 retries
I also applied the same fix to test_base_notif_clear_wait_ack in the scenario_base_notif_clear_on_disable scenario, which had the same unbounded loop. The retry ceiling is generous given the transport is fully in-memory and synchronous — any message enqueued by process_all_events is visible on the first dequeue attempt — but gives headroom if async delivery is introduced later.

lib/rpmi_context.c
icosta-sifive: Good nit, fixed — dropped the else and left the error path as a straight fall-through after the early return.

lib/rpmi_context.c
```diff
@@ -481,6 +599,23 @@ void rpmi_context_process_all_events(struct rpmi_context *cntx)
...
+	if (!cntx || !cntx->base_group) {
+		DPRINTF("%s: invalid parameters\n", __func__);
+		return;
+	}
+
+	base = (struct rpmi_base_group *)cntx->base_group->priv;
+
+	/* Set the pending flag for REQUEST_HANDLE_ERROR event */
+	rpmi_env_lock(cntx->base_group->lock);
+	base->pending_request_handle_err = true;
```
slingappa: I think the disable fix still misses the original disabled-window case. `REQ_STATE=0` now clears an already pending `pending_request_handle_err`, but `rpmi_context_base_request_handle_error()` still sets that pending bit unconditionally. Since `rpmi_base_process_events()` only suppresses delivery while `notif_enabled` is false, an ACK enqueue failure that happens after disable is stored and then delivered immediately after the AP re-enables notifications.

I reproduced this locally by moving the existing regression sequence to latch after the disable request is processed; the current no-spurious-notification check then fails with `spurious REQUEST_HANDLE_ERROR notification delivered after disable`. Could we either drop `REQUEST_HANDLE_ERROR` while notifications are disabled, or explicitly document/test that disabled-time events are queued for later delivery?

lib/rpmi_context.c
```diff
@@ -189,11 +256,52 @@ static enum rpmi_error rpmi_base_get_attributes(struct rpmi_service_group *group
...
+	rc = rpmi_transport_enqueue(trans, RPMI_QUEUE_P2A_REQ, notif_msg);
+	if (rc == RPMI_SUCCESS) {
+		/* Clear the pending flag on successful enqueue */
+		base->pending_request_handle_err = false;
+	} else if (rc == RPMI_ERR_IO) {
+		/* Queue is full, will retry later */
+		return RPMI_ERR_BUSY;
+	}
+
+	/* Other error occurred */
```
slingappa, line 293: After a successful `rpmi_transport_enqueue()`, this clears `pending_request_handle_err` but then falls through to the "failed to enqueue notification" `DPRINTF()` and returns `rc`. Runtime behavior is still success because `rc == RPMI_SUCCESS`, but debug builds will log a failure for every successful notification, and the following `return RPMI_SUCCESS;` is unreachable. Can we return immediately on success or move the failure log into the real non-IO error path?

lib/rpmi_context.c
icosta-sifive: Fixed by guarding the latch in rpmi_context_base_request_handle_error() with if (base->notif_enabled) — errors that arrive while notifications are disabled are dropped at latch time rather than queued. Added scenario_base_notif_drop_while_disabled which reproduces your sequence (enable → disable → trigger error while disabled → re-enable → process_all_events) and asserts P2A_REQ stays empty.

lib/rpmi_context.c
icosta-sifive: Fixed — added return RPMI_SUCCESS; immediately after clearing the pending flag so the DPRINTF("failed to enqueue...") is only reached on genuine non-IO errors. The unreachable return RPMI_SUCCESS at the end is removed.

slingappa (approved):
looks good to me.

icosta-sifive:
> looks good to me.

@slingappa Thanks....Still im unable to merge it if someone could it the button i appreciate

PR #107, opened 2026-07-08
lib: implement RAS_AGENT service group

icosta-sifive:
This PR implements the **RAS_AGENT** service group as specified in RPMI specification Section 4.12, providing services to enumerate error sources and retrieve their descriptors.

## Overview

The RAS_AGENT service group enables platform firmware to expose error sources to the application processor, allowing enumeration of error sources and retrieval of their descriptors in various formats (GHESv2/ACPI or implementation-specific).

### 1. Library Implementation (`lib/rpmi_service_group_ras.c`)

Implements all 4 services defined in the RPMI specification:

#### Services
- **RAS_ENABLE_NOTIFICATION** (0x01)
  - Returns `RPMI_ERR_NOTSUPP` (no events defined per specification)
  - Validates event ID and request state parameters

- **RAS_GET_NUM_ERR_SRCS** (0x02)
  - Returns total number of error sources in the system
  - No request parameters required

- **RAS_GET_ERR_SRCS_ID_LIST** (0x03)
  - Returns list of error source IDs with pagination support
  - Handles `START_INDEX` for partial list retrieval
  - Returns `REMAINING` and `RETURNED` counts
  - Validates start index bounds

- **RAS_GET_ERR_SRC_DESC** (0x04)
  - Returns error source descriptors with pagination support
  - Handles `BYTE_OFFSET` for partial descriptor retrieval
  - Supports GHESv2 (ACPI/APEI) format (`0x0`)
  - Supports implementation-specific format (`0xF`)
  - Validates error source ID and byte offset

#### Features
- **Pagination Support**: Automatically handles large data that exceeds transport slot size
- **Descriptor Format Support**: FLAGS[3:0] indicates format (GHESv2 or implementation-specific)
- **Platform HAL Interface**: Simple callback interface for platform descriptor retrieval
- **Endianness Handling**: Proper `rpmi_to_xe32()` usage for transport compatibility
- **Memory Safety**: Bounds checking, NULL pointer validation, overflow protection
- **Error Handling**: Comprehensive validation with appropriate error codes

### 2. Tests:

test/test_srvgrp_ras.c  - RAS test suite (new, 310 lines)
test/objects.mk  - Test build configuration

 ## Testing

All existing tests continue to pass. New notification functionality is verified through updated base service group tests.

## Compliance

Implements RAS_AGENT Service as specified in the RISC-V RPMI specification.

### DIFF
--- a/include/librpmi.h
+++ b/include/librpmi.h
@@ -79,6 +79,10 @@
 /** Data field size in bytes */
 #define RPMI_MSG_DATA_SIZE(__slot_size)		((__slot_size) - RPMI_MSG_HDR_SIZE)
 
+/** Read a uint32 field at index __idx from a service request data buffer */
+#define RPMI_REQ_U32(__trans, __data, __idx)	\
+	rpmi_to_xe32((__trans)->is_be, ((const rpmi_uint32_t *)(__data))[__idx])
+
 /** Minimum slot size in bytes */
 #define RPMI_SLOT_SIZE_MIN			(64)
 
@@ -347,6 +351,21 @@ enum rpmi_perf_service_id {
 	RPMI_PERF_SRV_GET_FAST_CHANNEL_REGION		= 0x09,
 	RPMI_PERF_SRV_GET_FAST_CHANNEL_ATTRIBUTES	= 0x0A,
 	RPMI_PERF_SRV_ID_MAX,
+};
+
+/** RPMI RAS Agent (RAS) ServiceGroup Service IDs */
+enum rpmi_ras_service_id {
+	RPMI_RAS_SRV_ENABLE_NOTIFICATION	= 0x01,
+	RPMI_RAS_SRV_GET_NUM_ERR_SRCS		= 0x02,
+	RPMI_RAS_SRV_GET_ERR_SRCS_ID_LIST	= 0x03,
+	RPMI_RAS_SRV_GET_ERR_SRC_DESC		= 0x04,
+	RPMI_RAS_SRV_ID_MAX,
+};
+
+/** RAS Error Source Descriptor Format Types */
+enum rpmi_ras_desc_format {
+	RPMI_RAS_DESC_FORMAT_GHESV2		= 0x0,
+	RPMI_RAS_DESC_FORMAT_IMPL_SPECIFIC	= 0xF,
 };
 
 /** RPMI Voltage (Volt) ServiceGroup Service IDs */
@@ -2121,7 +2140,82 @@ enum rpmi_error rpmi_mm_service_register(struct rpmi_service_group *group,
 
 /** @} */
 
-/****************************************************************************/
+/******************************************************************************/
+
+/**
+ * \defgroup LIBRPMI_RASAGENTSRVGRP_INTERFACE RPMI RAS Agent Service Group Library Interface
+ * @brief Global functions and data structures implemented by the RPMI library
+ * for RPMI RAS Agent service group.
+ * @{
+ */
+
+/** RAS error source descriptor data */
+struct rpmi_ras_err_src_desc {
+	/** Error source ID */
+	rpmi_uint32_t err_src_id;
+	/** Descriptor format (RPMI_RAS_DESC_FORMAT_*) */
+	rpmi_uint32_t desc_format;
+	/** Total descriptor size in bytes */
+	rpmi_uint32_t desc_size;
+	/** Pointer to descriptor data */
+	const rpmi_uint8_t *desc_data;
+};
+
+/** Platform specific RAS Agent operations */
+struct rpmi_ras_platform_ops {
+	/**
+	 * Optional: initialize platform state, called once during create.
+	 * Returns RPMI_SUCCESS on success, error code otherwise.
+	 */
+	enum rpmi_error (*init)(void *priv);
+
+	/**
+	 * Get total number of error sources.
+	 * Returns RPMI_SUCCESS on success, error code otherwise.
+	 */
+	enum rpmi_error (*get_num_error_sources)(void *priv,
+						 rpmi_uint32_t *num_err_srcs);
+
+	/**
+	 * Get up to @count error source IDs starting at @start_index into @ids.
+	 * Caller guarantees start_index + count <= total from get_num_error_sources.
+	 * Returns RPMI_SUCCESS on success, error code otherwise.
+	 */
+	enum rpmi_error (*get_err_src_id_list)(void *priv,
+					       rpmi_uint32_t start_index,
+					       rpmi_uint32_t count,
+					       rpmi_uint32_t *ids);
+
+	/**
+	 * Get error source descriptor for a given error source ID.
+	 * Returns RPMI_SUCCESS on success, error code otherwise.
+	 */
+	enum rpmi_error (*get_err_src_desc)(void *priv,
+					    rpmi_uint32_t err_src_id,
+					    struct rpmi_ras_err_src_desc *desc);
+};
+
+/**
+ * @brief Create a RAS Agent service group instance
+ *
+ * @param[in] ops		pointer to platform specific RAS operations
+ * @param[in] ops_priv		pointer to private data of platform operations
+ * @return pointer to RPMI service group instance upon success and NULL upon failure
+ */
+struct rpmi_service_group *
+rpmi_service_group_ras_create(const struct rpmi_ras_platform_ops *ops,
+			      void *ops_priv);
+
+/**
+ * @brief Destroy (or free) a RAS Agent service group instance
+ *
+ * @param[in] group	pointer to RPMI service group instance
+ */
+void rpmi_service_group_ras_destroy(struct rpmi_service_group *group);
+
+/** @} */
+
+/******************************************************************************/
 
 /**
  * \defgroup LIBRPMI_LOGGINGSRVGRP_INTERFACE RPMI Logging Service Group Library Interface
--- a/lib/rpmi_service_group_ras.c
+++ b/lib/rpmi_service_group_ras.c
@@ -1,0 +1,302 @@
+/*
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright (c) 2026 SiFive, Inc.
+ */
+
+#include <librpmi.h>
+
+/* Response field accessors (usable as lvalues) */
+#define RPMI_RAS_RESP_STATUS(__r)		((__r)[0])
+#define RPMI_RAS_RESP_FLAGS(__r)		((__r)[1])
+#define RPMI_RAS_RESP_REMAINING(__r)		((__r)[2])
+#define RPMI_RAS_RESP_RETURNED(__r)		((__r)[3])
+#define RPMI_RAS_RESP_DATA_PTR(__r)		(&(__r)[4])
+#define RPMI_RAS_RESP_ERR_SRC_ID(__r, __i)	((__r)[4 + (__i)])
+/* GET_NUM_ERR_SRCS response: status + count (no paginated header) */
+#define RPMI_RAS_RESP_NUM_ERR_SRCS(__r)		((__r)[1])
+
+/* Fill all four fixed fields of a paginated success response in one call */
+#define RPMI_RAS_PREPARE_RESPONSE(__t, __r, __flags, __remaining, __returned)	\
+	do {									\
+		RPMI_RAS_RESP_STATUS(__r)    = rpmi_to_xe32((__t)->is_be, RPMI_SUCCESS);	\
+		RPMI_RAS_RESP_FLAGS(__r)     = rpmi_to_xe32((__t)->is_be, (__flags));	\
+		RPMI_RAS_RESP_REMAINING(__r) = rpmi_to_xe32((__t)->is_be, (__remaining));\
+		RPMI_RAS_RESP_RETURNED(__r)  = rpmi_to_xe32((__t)->is_be, (__returned));	\
+	} while (0)
+
+/* Request field accessors */
+#define RPMI_RAS_REQ_NOTIF_STATE(__t, __d)	RPMI_REQ_U32(__t, __d, 1)
+#define RPMI_RAS_REQ_START_INDEX(__t, __d)	RPMI_REQ_U32(__t, __d, 0)
+#define RPMI_RAS_REQ_ERR_SRC_ID(__t, __d)	RPMI_REQ_U32(__t, __d, 0)
+#define RPMI_RAS_REQ_BYTE_OFFSET(__t, __d)	RPMI_REQ_U32(__t, __d, 1)
+
+/*
+ * Fixed response header: status, flags, remaining, returned (4 x uint32_t).
+ * Max IDs / bytes that fit in the variable payload of a paginated response.
+ */
+#define RPMI_RAS_MAX_ERR_SRCS_PER_RESP(__slot_size)	\
+	((RPMI_MSG_DATA_SIZE(__slot_size) - (4 * sizeof(rpmi_uint32_t))) /	\
+	 sizeof(rpmi_uint32_t))
+#define RPMI_RAS_MAX_DESC_BYTES_PER_RESP(__slot_size)	\
+	(RPMI_MSG_DATA_SIZE(__slot_size) - (4 * sizeof(rpmi_uint32_t)))
+
+#ifdef LIBRPMI_DEBUG
+#define DPRINTF(msg...)		rpmi_env_printf(msg)
+#else
+#define DPRINTF(msg...)
+#endif
+
+struct rpmi_ras_group {
+	const struct rpmi_ras_platform_ops *ops;
+	void *ops_priv;
+
+	struct rpmi_service_group group;
+};
+
+static enum rpmi_error rpmi_ras_enable_notification(struct rpmi_service_group *group,
+						    struct rpmi_service *service,
+						    struct rpmi_transport *trans,
+						    rpmi_uint16_t request_datalen,
+						    const rpmi_uint8_t *request_data,
+						    rpmi_uint16_t *response_datalen,
+						    rpmi_uint8_t *response_data)
+{
+	rpmi_uint32_t *resp = (void *)response_data;
+	rpmi_uint32_t state;
+
+	*response_datalen = sizeof(rpmi_uint32_t);
+
+	state = RPMI_RAS_REQ_NOTIF_STATE(trans, request_data);
+	if (state > 1) {
+		RPMI_RAS_RESP_STATUS(resp) = rpmi_to_xe32(trans->is_be, RPMI_ERR_INVALID_PARAM);
+		return RPMI_SUCCESS;
+	}
+
+	RPMI_RAS_RESP_STATUS(resp) = rpmi_to_xe32(trans->is_be, RPMI_ERR_NOTSUPP);
+	return RPMI_SUCCESS;
+}
+
+static enum rpmi_error rpmi_ras_get_num_err_srcs(struct rpmi_service_group *group,
+						 struct rpmi_service *service,
+						 struct rpmi_transport *trans,
+						 rpmi_uint16_t request_datalen,
+						 const rpmi_uint8_t *request_data,
+						 rpmi_uint16_t *response_datalen,
+						 rpmi_uint8_t *response_data)
+{
+	struct rpmi_ras_group *rasgrp = group->priv;
+	rpmi_uint32_t *resp = (void *)response_data;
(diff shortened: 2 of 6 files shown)


lib/rpmi_service_group_ras.c
```diff
@@ -0,0 +1,226 @@
+/*
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright (c) 2024 Ventana Micro Systems Inc.
```
hschauhan: Please change the copyright. There are other files with the same copyright please change them as well.

include/librpmi.h
```diff
@@ -2119,6 +2134,66 @@ enum rpmi_error rpmi_mm_service_register(struct rpmi_service_group *group,
...
+ * @param[in] ops		pointer to platform specific RAS operations
+ * @param[in] ops_priv		pointer to private data of platform operations
+ * @return pointer to RPMI service group instance upon success and NULL upon failure
+ */
+struct rpmi_service_group *
+rpmi_service_group_ras_create(rpmi_uint32_t num_err_srcs,
+			      const rpmi_uint32_t *err_src_ids,
+			      const struct rpmi_ras_platform_ops *ops,
+			      void *ops_priv);
+
```
hschauhan, line 2211: IMO, we should make this dynamic. This function should only take ops and ops_priv. ops should be extended to call "init" on the platform code and then ops should also have function pointers like "get_num_error_sources" and "get_err_src_ids".

include/librpmi.h
```diff
@@ -2119,6 +2134,66 @@ enum rpmi_error rpmi_mm_service_register(struct rpmi_service_group *group,
...
+/** Platform specific RAS Agent operations */
+struct rpmi_ras_platform_ops {
+	/**
+	 * Get error source descriptor for a given error source ID
+	 * Returns RPMI_SUCCESS on success, error code otherwise
+	 */
+	enum rpmi_error (*get_err_src_desc)(void *priv,
+					    rpmi_uint32_t err_src_id,
+					    struct rpmi_ras_err_src_desc *desc);
+};
```
hschauhan, line 2199: Please extend ops to accommodate initialization and service functions because this is layer that implements RPMI communication. Actual data is with the RAS agent and these services should be forwarded to the RAS agent. This also makes this layer dynamic.

hschauhan (commented):
Please check the review comment and propose new changes.

include/librpmi.h
icosta-sifive: Make Sense

include/librpmi.h
icosta-sifive: Done

slingappa (commented):
Review comments from local validation against the RPMI RAS_AGENT spec and targeted QEMU qtests.

lib/rpmi_service_group_ras.c
```diff
@@ -0,0 +1,260 @@
...
+		}
+	}
+
+	group = &rasgrp->group;
+	group->name = "ras_agent";
+	group->servicegroup_id = RPMI_SRVGRP_RAS_AGENT;
+	group->servicegroup_version =
+		RPMI_BASE_VERSION(RPMI_SPEC_VERSION_MAJOR, RPMI_SPEC_VERSION_MINOR);
+	/* Allowed only for M-mode RPMI context */
+	group->privilege_level_bitmap = RPMI_PRIVILEGE_M_MODE_MASK;
```
slingappa: Severity: MEDIUM

The RAS_AGENT service group should be available to both M-mode and S-mode contexts. The RPMI service-group table lists service group `0x000C RAS_AGENT` with allowed privilege levels `M-mode, S-mode`, but this code sets:

```c
group->privilege_level_bitmap = RPMI_PRIVILEGE_M_MODE_MASK;
```

That makes `rpmi_context_add_group()` fail for an S-mode RPMI context. I reproduced this with a QEMU qtest harness: `/riscv64/rpmi/ras/s-mode-access` gets `RPMI_ERR_DENIED` (`-4`) where the spec expects success. Please include `RPMI_PRIVILEGE_S_MODE_MASK` here as well.

lib/rpmi_service_group_ras.c
```diff
@@ -0,0 +1,260 @@
...
+
+	if (!rasgrp->ops->enable_notification) {
+		resp[0] = rpmi_to_xe32(trans->is_be, RPMI_ERR_NOTSUPP);
+		return RPMI_SUCCESS;
+	}
+
+	event_id = rpmi_to_xe32(trans->is_be, ((const rpmi_uint32_t *)request_data)[0]);
+	state = rpmi_to_xe32(trans->is_be, ((const rpmi_uint32_t *)request_data)[1]);
+
+	ret = rasgrp->ops->enable_notification(rasgrp->ops_priv, event_id,
```
slingappa: Severity: MEDIUM

`RAS_ENABLE_NOTIFICATION` should not be forwarded to a platform callback as a successful operation for this service group. Section 4.12.2 says RAS_AGENT does not support any notification events, and the common enable-notification response specifies `RPMI_ERR_NOT_SUPPORTED` for unsupported events (`RPMI_ERR_NOTSUPP` in this codebase) and `RPMI_ERR_INVALID_PARAM` for invalid request state values.

With a callback that returns `RPMI_SUCCESS`, the current code returns success for event 0/state 1. The QEMU qtest `/riscv64/rpmi/ras/no-notification-events` reproduces this as `(0 == -2)`. Please make this service reject all RAS notification events consistently, or at least validate `REQ_STATE` and never report success for an unsupported event. If success is ever possible, the response also needs to include `CURRENT_STATE` per the common response table.

lib/rpmi_service_group_ras.c
```diff
@@ -0,0 +1,260 @@
...
+		*response_datalen = sizeof(rpmi_uint32_t);
+		return RPMI_SUCCESS;
+	}
+
+	/* Calculate how many error source IDs we can fit in response */
+	max_err_srcs = (RPMI_MSG_DATA_SIZE(trans->slot_size) - (4 * sizeof(rpmi_uint32_t))) / sizeof(rpmi_uint32_t);
+	remaining = num_err_srcs - start_index;
+	returned = (remaining > max_err_srcs) ? max_err_srcs : remaining;
+
+	ret = rasgrp->ops->get_err_src_id_list(rasgrp->ops_priv, start_index,
```
slingappa, line 140: Severity: MEDIUM

The returned `RAS_ERR_SRC_ID[N]` array is copied directly into the response and never converted to the transport endianness. The scalar response words are converted with `rpmi_to_xe32()`, but the IDs written by `get_err_src_id_list()` remain host-endian in `resp[4..]`.

The QEMU qtest `/riscv64/rpmi/ras/id-list-big-endian` reproduced this: the response contained `0x01020304` where a big-endian transport expected `0x04030201`. Please loop over the returned IDs after the platform callback and convert each `resp[4 + i]` with `rpmi_to_xe32(trans->is_be, ...)`, matching the pattern used by other list-returning service groups.

lib/rpmi_service_group_ras.c
```diff
@@ -0,0 +1,260 @@
...
+	}
+
+	if (start_index >= num_err_srcs) {
+		resp[0] = rpmi_to_xe32(trans->is_be, RPMI_ERR_INVALID_PARAM);
+		*response_datalen = sizeof(rpmi_uint32_t);
+		return RPMI_SUCCESS;
+	}
+
+	/* Calculate how many error source IDs we can fit in response */
+	max_err_srcs = (RPMI_MSG_DATA_SIZE(trans->slot_size) - (4 * sizeof(rpmi_uint32_t))) / sizeof(rpmi_uint32_t);
```
slingappa: Severity: LOW

`checkpatch.pl --no-tree` reports this line exceeds 100 columns:

```c
max_err_srcs = (RPMI_MSG_DATA_SIZE(trans->slot_size) - (4 * sizeof(rpmi_uint32_t))) / sizeof(rpmi_uint32_t);
```

Please wrap this calculation to match the surrounding style.

test/test_srvgrp_ras.c
```diff
@@ -0,0 +1,349 @@
...
+	    err_src_id != TEST_RAS_ERR_SRC_ID_1 &&
+	    err_src_id != TEST_RAS_ERR_SRC_ID_2)
+		return RPMI_ERR_INVALID_PARAM;
+
+	desc->err_src_id = err_src_id;
+	desc->desc_format = RPMI_RAS_DESC_FORMAT_GHESV2;
+	desc->desc_size = sizeof(test_ghesv2_desc);
+	desc->desc_data = test_ghesv2_desc;
+
+	printf("platform callback: returning descriptor for error source ID 0x%x (format=%d, size=%d bytes)\n",
```
slingappa: Severity: LOW

`checkpatch.pl --no-tree` reports this debug `printf()` exceeds 100 columns:

```c
printf("platform callback: returning descriptor for error source ID 0x%x (format=%d, size=%d bytes)\n",
```

Please wrap or shorten the format string.

test/test_srvgrp_ras.c
```diff
@@ -0,0 +1,349 @@
...
+		{
+			.name = "GET ERR SRC DESC TEST (invalid byte offset)",
+			.attrs = {
+				.servicegroup_id = RPMI_SRVGRP_RAS_AGENT,
+				.service_id = RPMI_RAS_SRV_GET_ERR_SRC_DESC,
+				.flags = RPMI_MSG_NORMAL_REQUEST,
+				.request_data = get_err_src_desc_reqdata_invalid_offset,
+				.request_data_len = sizeof(get_err_src_desc_reqdata_invalid_offset),
+				.expected_data = get_err_src_desc_expdata_invalid_offset,
+				.expected_data_len = sizeof(get_err_src_desc_expdata_invalid_offset),
```
slingappa: Severity: LOW

`checkpatch.pl --no-tree` reports this initializer line exceeds 100 columns:

```c
.expected_data_len = sizeof(get_err_src_desc_expdata_invalid_offset),
```

Please wrap it before merge.

slingappa (commented):
Review comments from local validation against the RPMI RAS_AGENT spec and targeted QEMU qtests.

lib/rpmi_service_group_ras.c
icosta-sifive: Fixed — RAS_AGENT (0x000C) is listed as M-mode + S-mode in the spec. Removed the incorrect comment and added RPMI_PRIVILEGE_S_MODE_MASK.

lib/rpmi_service_group_ras.c
icosta-sifive: Fixed. Since Section 4.12.2 specifies no notification events for RAS_AGENT, the handler now validates REQ_STATE (returns RPMI_ERR_INVALID_PARAM for values outside 0–1) and unconditionally returns RPMI_ERR_NOTSUPP — no callback forwarding. Also removed the now-dead enable_notification ops callback from struct rpmi_ras_group_ops.

lib/rpmi_service_group_ras.c
icosta-sifive: Fixed. After get_err_src_id_list() returns, each resp[4 + i] is now passed through rpmi_to_xe32(trans->is_be, ...) before the response is sent, matching the pattern used in the voltage and clock groups.

lib/rpmi_service_group_ras.c
icosta-sifive: Fixed — wrapped the line at the subtraction operator to stay within 100 columns.

test/test_srvgrp_ras.c
icosta-sifive: Fixed — split the format string literal across two lines to stay within 100 columns.

test/test_srvgrp_ras.c
icosta-sifive: Fixed — wrapped .expected_data_len initializer to stay within 100 columns.

lib/rpmi_service_group_ras.c
```diff
@@ -0,0 +1,258 @@
...
+	}
+
+	if (start_index >= num_err_srcs) {
+		resp[0] = rpmi_to_xe32(trans->is_be, RPMI_ERR_INVALID_PARAM);
+		*response_datalen = sizeof(rpmi_uint32_t);
+		return RPMI_SUCCESS;
+	}
+
+	/* Calculate how many error source IDs we can fit in response */
+	max_err_srcs = (RPMI_MSG_DATA_SIZE(trans->slot_size) -
```
hschauhan: I think we should have a clean macro which does this math.

lib/rpmi_service_group_ras.c
```diff
@@ -0,0 +1,258 @@
...
+{
+	struct rpmi_ras_group *rasgrp = group->priv;
+	struct rpmi_ras_err_src_desc desc;
+	rpmi_uint32_t err_src_id, byte_offset, remaining, returned;
+	rpmi_uint32_t *resp = (void *)response_data;
+	rpmi_uint8_t *desc_data = (rpmi_uint8_t *)&resp[4];
+	rpmi_uint16_t max_desc_bytes;
+	enum rpmi_error ret;
+
+	err_src_id = rpmi_to_xe32(trans->is_be, ((const rpmi_uint32_t *)request_data)[0]);
```
hschauhan: Access macros like GET_ERR_SRC_ID(trans, request-data) and similar macros at other places.

lib/rpmi_service_group_ras.c
hschauhan: I am uncomfortable with direct index manipulation of resp. These are prone to error. Please add access macros that can gate the access and manipulation of the array.

lib/rpmi_service_group_ras.c
```diff
@@ -0,0 +1,258 @@
...
+	ret = rasgrp->ops->get_err_src_id_list(rasgrp->ops_priv, start_index,
+					       returned, &resp[4]);
+	if (ret != RPMI_SUCCESS) {
+		resp[0] = rpmi_to_xe32(trans->is_be, ret);
+		*response_datalen = sizeof(rpmi_uint32_t);
+		return RPMI_SUCCESS;
+	}
+
+	for (rpmi_uint32_t i = 0; i < returned; i++)
+		resp[4 + i] = rpmi_to_xe32(trans->is_be, resp[4 + i]);
```
hschauhan: Same here. Gate these direct manipulation, please add and use GET/SET macros

lib/rpmi_service_group_ras.c
```diff
@@ -0,0 +1,258 @@
...
+
+	resp[0] = rpmi_to_xe32(trans->is_be, RPMI_SUCCESS);
+	resp[1] = 0; /* FLAGS - reserved */
+	resp[2] = rpmi_to_xe32(trans->is_be, remaining - returned);
+	resp[3] = rpmi_to_xe32(trans->is_be, returned);
```
hschauhan: #define PREPARE_RESPONSE(....)

lib/rpmi_service_group_ras.c
```diff
@@ -0,0 +1,258 @@
...
+
+	/* Calculate how many bytes we can fit in response */
+	max_desc_bytes = RPMI_MSG_DATA_SIZE(trans->slot_size) - (4 * sizeof(rpmi_uint32_t));
+	remaining = desc.desc_size - byte_offset;
+	returned = (remaining > max_desc_bytes) ? max_desc_bytes : remaining;
+
+	resp[0] = rpmi_to_xe32(trans->is_be, RPMI_SUCCESS);
+	resp[1] = rpmi_to_xe32(trans->is_be, desc.desc_format & 0xF); /* FLAGS[3:0] = format */
+	resp[2] = rpmi_to_xe32(trans->is_be, remaining - returned);
+	resp[3] = rpmi_to_xe32(trans->is_be, returned);
```
hschauhan: PREPARE_RESPONSE()

hschauhan (changes requested):

lib/rpmi_service_group_ras.c
icosta-sifive: Good Point, I'ill introduce a macro for this

lib/rpmi_service_group_ras.c
icosta-sifive: Agreed. The same verbose pattern appears across many service groups, so I'ill add a general helper to librpmi.h

lib/rpmi_service_group_ras.c
icosta-sifive: Done. NO vare numeric indices remain in any of the four service handlers

lib/rpmi_service_group_ras.c
icosta-sifive: Added RPMI_RAS_PREPARE_RESPONSE, which writes all four fixed response fields in one call.

hschauhan (approved):

lib/rpmi_service_group_ras.c
```diff
@@ -0,0 +1,287 @@
...
+	rpmi_uint32_t *resp = (void *)response_data;
+	rpmi_uint8_t *desc_data = (rpmi_uint8_t *)RPMI_RAS_RESP_DATA_PTR(resp);
+	rpmi_uint16_t max_desc_bytes;
+	enum rpmi_error ret;
+
+	err_src_id  = RPMI_RAS_REQ_ERR_SRC_ID(trans, request_data);
+	byte_offset = RPMI_RAS_REQ_BYTE_OFFSET(trans, request_data);
+
+	/* Get descriptor from platform; platform validates err_src_id */
+	ret = rasgrp->ops->get_err_src_desc(rasgrp->ops_priv, err_src_id, &desc);
```
slingappa, line 178: Severity: MEDIUM

`get_err_src_desc()` trusts every field returned by the platform callback before building the wire response. The RAS spec defines only GHESv2 (`0`) and implementation-specific (`0xF`) descriptor formats in `FLAGS[3:0]`; reserved/high-bit values should not be silently normalized by `desc.desc_format & 0xF`. The same path also copies from `desc.desc_data + byte_offset` without checking that `desc.desc_data` is non-NULL when bytes are returned.

Please validate the callback output before `RPMI_RAS_PREPARE_RESPONSE()`: accept only `RPMI_RAS_DESC_FORMAT_GHESV2` or `RPMI_RAS_DESC_FORMAT_IMPL_SPECIFIC`, reject reserved/high-bit formats, and reject non-zero descriptor sizes without a data pointer.

test/test_srvgrp_ras.c
```diff
@@ -0,0 +1,350 @@
...
+	if (!grp) {
+		printf("failed to create RAS agent service group\n");
+		return RPMI_ERR_FAILED;
+	}
+
+	rpmi_context_add_group(scene->cntx, grp);
+	return 0;
+}
+
+static struct rpmi_test_scenario scenario_ras_default = {
```
slingappa, line 207: Severity: MEDIUM

The code now fixes the earlier S-mode access and big-endian ID-list issues, but the new RAS test scenario still only runs through `test_scenario_default_init()`, which creates an M-mode context, and the shared-memory transport defaults to little-endian. That means both previously reviewed regressions could come back while `test_srvgrp_ras` still passes.

Please add focused regression coverage for an S-mode RAS context and for a big-endian `RAS_GET_ERR_SRCS_ID_LIST` response that verifies each returned `RAS_ERR_SRC_ID[N]` is encoded through `rpmi_to_xe32()`.

icosta-sifive: @slingappa Done. Added two new test scenarios:

scenario_ras_smode — creates an S-mode context and runs ENABLE_NOTIFICATION, GET_NUM_ERR_SRCS, GET_ERR_SRCS_ID_LIST and GET_ERR_SRC_DESC through it to catch any future S-mode access regression.
scenario_ras_be — runs GET_ERR_SRCS_ID_LIST over a big-endian transport and verifies each returned ID is encoded via rpmi_to_xe32() using a custom init_expected_data callback that builds the expected data with rpmi_to_xe32(true, ...).
While implementing the big-endian scenario I found a pre-existing bug in rpmi_transport_enqueue/rpmi_transport_dequeue: both were using rpmi_to_xe32 to byte-swap servicegroup_id and datalen, which are rpmi_uint16_t fields. On a big-endian transport bswap32(0x000C) = 0x0C000000, which truncates back to uint16_t as 0x0000, causing the context to silently drop every request. Fixed by replacing those conversions with rpmi_to_xe16. This is a no-op for little-endian transports, so no existing tests are affected.

slingappa (approved):
Looks good to me.

icosta-sifive:
> looks good to me.

@slingappa Thanks....Same here im unable to merge it if someone could it the button i appreciate

pathakraul:
Looks good to me

PR #108, opened 2026-07-10
lib: implement REQUEST_FORWARD service group

icosta-sifive:
This PR implements the REQUEST_FORWARD service group as specified in RPMI specification Section 4.13, providing services to retrieve and process RPMI request messages forwarded by the platform microcontroller.

### Overview
The REQUEST_FORWARD service group enables application processors to retrieve and process RPMI request messages that are forwarded by platform firmware from other RPMI clients. This supports use-cases like forwarding RPMI messages between system-level partitions/domains using the SBI MPXY extension.

### Changes

**1. Library Implementation ( lib/rpmi_service_group_reqfwd.c )**

Implements all 3 services defined in the RPMI specification:

#### Services
- **REQFWD_ENABLE_NOTIFICATION** (0x01)
  - Enable, disable, or query NEW_MESSAGE event notifications
  - Validates event ID and request state parameters
  - Returns current notification state

- **REQFWD_RETRIEVE_CURRENT_MESSAGE** (0x02)
  - Retrieves the current (oldest) forwarded RPMI message from platform queue
  - Handles START_INDEX for pagination of large messages
  - Returns REMAINING and RETURNED byte counts
  - Validates start index bounds

- **REQFWD_COMPLETE_CURRENT_MESSAGE** (0x03)
  - Completes processing of the current forwarded message
  - Accepts optional response data to send back to original requester
  - Returns NUM_MESSAGES (count of remaining messages in queue)
  - Validates that a message was previously retrieved

**Features**
- **Pagination Support**: Automatically handles large messages exceeding transport slot size
- **State Management**: Prevents completing a message without prior retrieval
- **Platform HAL Interface**: Clean callback interface for message retrieval and completion
- **Endianness Handling**: Proper  rpmi_to_xe32()  usage for transport compatibility
- **Memory Safety**: Bounds checking, NULL pointer validation, overflow protection
- **Error Handling**: Comprehensive validation with appropriate error codes

**2. Tests:**

test/test_srvgrp_reqfwd.c  - REQUEST_FORWARD test suite (new, 173 lines)
test/objects.mk  - Test build configuration

### Testing

All existing tests continue to pass. New REQUEST_FORWARD functionality is verified through the test suite.

Compliance
Implements REQUEST_FORWARD Service Group (0x000D) as specified in the RISC-V RPMI specification Section 4.13.

### DIFF
--- a/lib/rpmi_service_group_reqfwd.c
+++ b/lib/rpmi_service_group_reqfwd.c
@@ -1,0 +1,351 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2026 SiFive, Inc.
+ */
+
+#include <librpmi.h>
+
+#ifdef LIBRPMI_DEBUG
+#define DPRINTF(msg...)		rpmi_env_printf(msg)
+#else
+#define DPRINTF(msg...)
+#endif
+
+/** Request Forward service group private data */
+struct reqfwd_service_group {
+	const struct rpmi_reqfwd_platform_ops	*ops;
+	void					*ops_priv;
+	rpmi_bool_t				message_retrieved;
+	rpmi_bool_t				notif_enabled;
+	rpmi_bool_t				pending_new_message;
+	rpmi_uint32_t				pending_event_data_len;
+	struct rpmi_transport			*trans;
+	struct rpmi_message			*notif_msg;
+
+	struct rpmi_service_group		group;
+};
+
+static enum rpmi_error rpmi_reqfwd_enable_notification(struct rpmi_service_group *group,
+						       struct rpmi_service *service,
+						       struct rpmi_transport *trans,
+						       rpmi_uint16_t request_datalen,
+						       const rpmi_uint8_t *request_data,
+						       rpmi_uint16_t *response_datalen,
+						       rpmi_uint8_t *response_data)
+{
+	rpmi_uint32_t *resp = (rpmi_uint32_t *)response_data;
+	const rpmi_uint32_t *req = (const rpmi_uint32_t *)request_data;
+	struct reqfwd_service_group *rgrp = group->priv;
+	rpmi_uint32_t event_id, req_state;
+
+	/* Notifications require a P2A channel to deliver events */
+	if (!trans->is_p2a_channel) {
+		resp[0] = rpmi_to_xe32(trans->is_be, RPMI_ERR_NOTSUPP);
+		*response_datalen = sizeof(rpmi_uint32_t);
+		return RPMI_SUCCESS;
+	}
+
+	event_id = rpmi_to_xe32(trans->is_be, req[0]);
+	req_state = rpmi_to_xe32(trans->is_be, req[1]);
+
+	DPRINTF("%s: event_id=%d req_state=%d\n", __func__, event_id, req_state);
+
+	/* Validate event ID */
+	if (event_id != RPMI_REQFWD_EVENT_NEW_MESSAGE) {
+		resp[0] = rpmi_to_xe32(trans->is_be, RPMI_ERR_INVALID_PARAM);
+		*response_datalen = sizeof(rpmi_uint32_t);
+		return RPMI_SUCCESS;
+	}
+
+	/* Validate and process request state */
+	switch (req_state) {
+	case 0: /* Disable */
+		rgrp->notif_enabled = false;
+		rgrp->pending_new_message = false;
+		break;
+	case 1: /* Enable */
+		rgrp->notif_enabled = true;
+		break;
+	case 2: /* Query current state */
+		break;
+	default:
+		resp[0] = rpmi_to_xe32(trans->is_be, RPMI_ERR_INVALID_PARAM);
+		*response_datalen = sizeof(rpmi_uint32_t);
+		return RPMI_SUCCESS;
+	}
+
+	/* Return current state */
+	resp[0] = rpmi_to_xe32(trans->is_be, RPMI_SUCCESS);
+	resp[1] = rpmi_to_xe32(trans->is_be, rgrp->notif_enabled ? 1U : 0U);
+	*response_datalen = 2 * sizeof(rpmi_uint32_t);
+
+	return RPMI_SUCCESS;
+}
+
+static enum rpmi_error rpmi_reqfwd_retrieve_current_message(struct rpmi_service_group *group,
+							    struct rpmi_service *service,
+							    struct rpmi_transport *trans,
+							    rpmi_uint16_t request_datalen,
+							    const rpmi_uint8_t *request_data,
+							    rpmi_uint16_t *response_datalen,
+							    rpmi_uint8_t *response_data)
+{
+	rpmi_uint32_t *resp = (rpmi_uint32_t *)response_data;
+	const rpmi_uint32_t *req = (const rpmi_uint32_t *)request_data;
+	struct reqfwd_service_group *rgrp = group->priv;
+	struct rpmi_reqfwd_message msg;
+	rpmi_uint32_t start_index, max_data, remaining, returned;
+	enum rpmi_error ret;
+
+	start_index = rpmi_to_xe32(trans->is_be, req[0]);
+
+	DPRINTF("%s: start_index=%d\n", __func__, start_index);
+
+	/* Get the current forwarded message from platform */
+	ret = rgrp->ops->get_current_message(rgrp->ops_priv, &msg);
+	if (ret != RPMI_SUCCESS) {
+		resp[0] = rpmi_to_xe32(trans->is_be, ret);
+		*response_datalen = sizeof(rpmi_uint32_t);
+		return RPMI_SUCCESS;
+	}
+
+	/* Check if we have a message */
+	if (msg.msg_size == 0 || !msg.msg_data) {
+		resp[0] = rpmi_to_xe32(trans->is_be, RPMI_ERR_NO_DATA);
+		*response_datalen = sizeof(rpmi_uint32_t);
+		return RPMI_SUCCESS;
+	}
+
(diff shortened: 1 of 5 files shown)


slingappa (commented):
Review comments from local qtest/checkpatch pass. Excluding SPDX/checkpatch warnings that appear to match existing repo style.

lib/rpmi_service_group_reqfwd.c
```diff
@@ -0,0 +1,276 @@
...
+		return;
+	}
+
+	rpmi_env_free_lock(group->lock);
+	rpmi_env_free(group->priv);
+
+	DPRINTF("%s: group destroyed\n", __func__);
+}
+
+enum rpmi_error rpmi_reqfwd_trigger_new_message(struct rpmi_service_group *group,
```
slingappa, line 316: `rpmi_reqfwd_trigger_new_message()` currently returns `RPMI_SUCCESS` without actually queueing a `NEW_MESSAGE` notification. After notification is enabled it discards `event_data_len`/`event_data`, logs that P2A notification queuing is not supported, and returns success. The service group also does not install a `process_events` hook that could send the notification later.

I reproduced this through a real QEMU virt RPMI shared-memory path by adding local QEMU test glue to register REQUEST_FORWARD and call this trigger API after processing the doorbell. `Base.PROBE REQUEST_FORWARD` returned `0x00010000`, and retrieve/complete paths worked, but after `ENABLE_NOTIFICATION` no P2A request notification was produced:

```text
REQ REQFWD.ENABLE_NOTIFICATION
BUG missing_new_message_notification_after_trigger
RPMI_QTEST_RESULT FAIL missing_new_message_notification_after_trigger
```

Please either implement the P2A notification delivery for `RPMI_REQFWD_EVENT_NEW_MESSAGE` or return an error when delivery is unsupported. This should also have a test that enables the event, triggers a new message, and verifies the notification contents.

test/test_srvgrp_reqfwd.c
```diff
@@ -0,0 +1,348 @@
...
+};
+
+/* Mock forwarded message data */
+static rpmi_uint8_t mock_forwarded_msg[] = {
+	0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08,
+	0x09, 0x0a, 0x0b, 0x0c, 0x0d, 0x0e, 0x0f, 0x10,
+};
+
+/* Mock state */
+static rpmi_bool_t mock_has_message = 0;
```
slingappa: Minor checker cleanup before respinning: `git diff --check` reports a trailing blank line at EOF in `test/test_srvgrp_reqfwd.c`, and checkpatch reports explicit static zero initializers for `mock_has_message` and `mock_num_pending`. Please drop the trailing blank line and initialize those statics implicitly.

lib/rpmi_service_group_reqfwd.c
icosta-sifive: Thanks @slingappa for the detailed report
The issue is fixed in the latest commit. Here's what changed:

lib/rpmi_service_group_reqfwd.c

rpmi_reqfwd_trigger_new_message now returns RPMI_ERR_NOTSUPP immediately when the transport has no P2A channel, instead of silently succeeding. When P2A is available, it builds the notification message in a pre-allocated buffer (event_id + any event_data bytes, bounded by slot size) and sets pending_new_message = true.
Added rpmi_reqfwd_process_events: called by rpmi_context_process_all_events, it delivers the pending notification to RPMI_QUEUE_P2A_REQ when evt_notify_state is enabled. Returns RPMI_ERR_BUSY on queue-full (RPMI_ERR_IO) so delivery is retried on the next event processing cycle.
rpmi_service_group_reqfwd_create now takes a struct rpmi_transport *trans parameter (required for P2A channel availability check and notification enqueue), allocates the notif_msg buffer, and registers the process_events hook.
test/test_srvgrp_reqfwd.c

Added two test cases covering the reported scenario:

Re-enables RPMI_REQFWD_EVENT_NEW_MESSAGE notification
Calls rpmi_reqfwd_trigger_new_message, lets rpmi_context_process_all_events run, then dequeues from RPMI_QUEUE_P2A_REQ and verifies the notification contains the correct event ID

lib/rpmi_service_group_reqfwd.c
```diff
@@ -0,0 +1,336 @@
...
+
+	if (!group || group->servicegroup_id != RPMI_SRVGRP_REQUEST_FORWARD)
+		return RPMI_ERR_INVALID_PARAM;
+
+	rgrp = group->priv;
+
+	if (!rgrp->trans->is_p2a_channel)
+		return RPMI_ERR_NOTSUPP;
+
+	(void)event_data_len;
```
slingappa: The new notification path now delivers a P2A `NEW_MESSAGE` event, so the previous missing-delivery issue looks fixed. However, this trigger API still drops the documented event payload: `include/librpmi.h` says `event_data_len` is the length of the event data, specifically the first N bytes of the message, but `rpmi_reqfwd_trigger_new_message()` discards both `event_data_len` and `event_data`. `rpmi_reqfwd_process_events()` then always sends a 4-byte notification containing only the event ID.

That means a platform caller can pass nonzero event data and get `RPMI_SUCCESS`, but the AP never receives the documented bytes. Please either implement the payload path by storing a bounded copy of `event_data`, validating `event_data_len == 0 || event_data != NULL`, and setting notification `datalen` to `sizeof(event_id) + event_data_len`, or simplify the public API/docs if Request Forward notifications are intentionally event-ID-only. It would be good to add a test that triggers with nonzero event data and verifies the P2A notification contents.

slingappa (commented):
Reviewed the latest revision. The earlier missing P2A notification delivery and style cleanup comments look addressed; I found one remaining API/behavior mismatch below.

lib/rpmi_service_group_reqfwd.c
icosta-sifive: Fixed — trigger_new_message now validates event_data != NULL when event_data_len > 0, clamps the length to fit the slot, and pre-populates the notification buffer with event_id || event_data. process_events sets datalen = sizeof(event_id) + pending_event_data_len. A new test covers the nonzero-payload case end-to-end.

slingappa: Thanks, the payload-copy part is addressed now: `event_data` is validated, bounded, copied into the notification buffer, and the new test covers a nonzero payload. I think there is still a wire-format issue in the same path, though. The RPMI notification format defines the first event word as `EVENT_HDR`, with `EVENT_ID` in bits `[23:16]` and `EVENT_DATALEN` in bits `[15:0]`; `REQFWD_NEW_MESSAGE` then uses that `EVENT_DATALEN` to describe the following bytes.

This code still writes only `RPMI_REQFWD_EVENT_NEW_MESSAGE` into the first word. That emits `0x00000001` for a no-payload notification instead of `0x00010000`, and with a 4-byte payload it still emits `0x00000001` instead of `0x00010004`. A spec-compliant AP parser would decode that as event ID 0 with an invalid/nonzero data length. Please encode the event header as `(RPMI_REQFWD_EVENT_NEW_MESSAGE << 16) | event_data_len`, convert that word with `rpmi_to_xe32()`, and update the tests to expect the encoded event header. It would also be worth rejecting or rounding down non-4-byte `event_data_len` values, since notification event data length must be 4-byte aligned.

lib/rpmi_service_group_reqfwd.c
```diff
@@ -0,0 +1,349 @@
...
+	if (event_data_len > max_event_data)
+		event_data_len = max_event_data;
+
+	/* Pre-populate the notification payload: event_id followed by event data */
+	((rpmi_uint32_t *)rgrp->notif_msg->data)[0] =
+		rpmi_to_xe32(rgrp->trans->is_be, RPMI_REQFWD_EVENT_NEW_MESSAGE);
+	if (event_data_len)
+		rpmi_env_memcpy((rpmi_uint8_t *)rgrp->notif_msg->data +
+				sizeof(rpmi_uint32_t), event_data, event_data_len);
+	rgrp->pending_event_data_len = event_data_len;
```
slingappa, line 345: `rpmi_reqfwd_trigger_new_message()` updates the pending notification state and `notif_msg` buffer without taking `group->lock`, but `rpmi_reqfwd_process_events()` reads the same state under that lock via `rpmi_context_process_all_events()` / `rpmi_context_process_group_events()`. Since this trigger is exported as a platform-facing API, a real integration can call it from a different event path while RPMI event processing is running.

That can race and enqueue a mixed payload/length, stale data, or lose a pending notification. Please protect the payload copy, `pending_event_data_len`, and `pending_new_message` updates with `group->lock` as the existing `rpmi_service_group_sysmsi_inject()` helper does for pending MSI state, or document that callers must already hold the service-group lock and provide a locked wrapper.

icosta-sifive: Agree — the first event word was written as RPMI_REQFWD_EVENT_NEW_MESSAGE (0x01) instead of the packed EVENT_HDR field.

lib/rpmi_service_group_reqfwd.c
icosta-sifive: Agree — the first event word was written as RPMI_REQFWD_EVENT_NEW_MESSAGE (0x01) instead of the packed EVENT_HDR field.

lib/rpmi_service_group_reqfwd.c
icosta-sifive: Fixed. rpmi_reqfwd_trigger_new_message() now acquires group->lock before writing the event header, copying the payload, and setting pending_event_data_len / pending_new_message, and releases it after, matching the pattern used by rpmi_service_group_sysmsi_inject().

slingappa (approved):
Looks good to me.

PR #109, opened 2026-07-13
test: pass NULL as ops_priv in sysreset test scenarios

icosta-sifive:
The ops_priv parameter in rpmi_service_group_sysreset_create() is intended for private data that the platform (low-level layer) wants the RPMI infrastructure to pass back to its platform callbacks. Passing scene->cntx (an rpmi_context object belonging to the RPMI middle layer) was semantically incorrect.

The test platform callbacks do not use the priv argument at all — they track state through static globals — so the correct value is NULL.


pathakraul:
Looks good to me

PR #106, opened 2026-07-07
Bug: lib: fix system suspend service group definition and add SYSTEM_SUSPEND test

icosta-sifive:
This PR fixes a critical bug in the system suspend service group and adds comprehensive test coverage to prevent similar issues in the future.

## Overview

Fixes a copy-paste error in the system suspend service group definition where system reset constants were incorrectly used, causing the entire service group to be non-functional. Adds a complete test suite following the same architecture as existing service group tests.

### Bug Fix: System Suspend Service Group Definition (`lib/rpmi_service_group_syssusp.c`)

Fixed critical copy-paste error in the service array definition:

- **Before:** 
  - Struct name: `rpmi_sys_services` 
  - Service constants: `RPMI_SYSRST_SRV_*` (system reset)
- **After:** 
  - Struct name: `rpmi_syssusp_services`
  - Service constants: `RPMI_SYSSUSP_SRV_*` (system suspend)

This bug rendered the entire system suspend service group non-functional as it was referencing incorrect service IDs from the system reset service group.

### Test Suite: System Suspend Service Group (`test/test_srvgrp_syssusp.c`)

Added comprehensive test coverage for the system suspend service group following the same patterns established by `test_srvgrp_sysreset.c`:

#### Test Cases (6 total)
- **ENABLE_NOTIFICATION** - Verifies NOTSUPP status (notifications not currently supported)
- **NOTIFICATION_STATUS** - Verifies NOTSUPP status for status queries
- **GET_ATTRIBUTES (supported type)** - Validates SUSPEND_TO_RAM type attributes
- **GET_ATTRIBUTES (unsupported type)** - Verifies proper error handling for invalid types
- **SYSTEM_SUSPEND (supported type)** - Validates suspend execution and platform callbacks
- **SYSTEM_SUSPEND (unsupported type)** - Verifies error handling without triggering callbacks

## Testing

- [x] `make` - builds successfully without warnings
- [x] `make check` - all tests pass
- [x] `make LIBRPMI_TEST=y` - builds with tests successfully

Executing System Suspend Service Group test scenario :
-------------------------------------------------
TEST: ENABLE NOTIFICATION TEST (notifications not supported)  : Succeeded!
TEST: NOTIFICATION STATUS TEST (notifications not supported)  : Succeeded!
TEST: GET_ATTRIBUTES TEST (for supported suspend type)        : Succeeded!
TEST: GET_ATTRIBUTES TEST (for unsupported suspend type)      : Succeeded!
[...]

PR #111, opened 2026-07-22
build: fix Makefile portability and test stability across platforms

icosta-sifive:
**Makefile bug fix**

1. GNU Make treats `#` as a comment character inside `$(shell ...)`, silently truncating `#define` in the sed pattern and leaving `$(strip` unterminated. A secondary issue: `\)` in the sed capture group was counted by Make as a bare ) closing `$(shell` prematurely.
Fixed by replacing sed with `grep | awk` — no `#` in the Make expression.

2. The compile_so function used  `-Wl,-soname`,... unconditionally — a Linux-only linker flag. Apple's ld uses -install_name instead and does not recognise -soname, causing the shared library build to fail on any macOS version.
Fixed by making the flag conditional on the platform, mirroring the existing Darwin check already present for ELFFLAGS

3. Test ELFs were linked with -static on Linux, which forces the entire binary to be statically linked — including libc. This requires libc.a to be present on the build host, which is not guaranteed (it ships separately as glibc-static on RHEL/Fedora and is often absent on servers and CI environments).
macOS was already using a better approach: passing $(build_dir)/librpmi.a directly to the linker, which statically links only librpmi while leaving libc dynamically linked. The same approach now applies to Linux, making both platforms consistent and removing the libc.a dependency entirely.

**Test fix**

test_syssusp_scenario_init declared test_hartid_array as a local stack variable and passed it to rpmi_hsm_create, which stores the pointer directly (hsm->leaf.hart_ids = hart_ids) without copying. Once the function returned, the HSM held a dangling pointer to deallocated stack memory.

At test runtime, rpmi_hsm_hart_id2index reads through that pointer to match the hart ID. On Linux x86_64 the stack frame happened to remain intact, so the stale value was still readable and the test passed. On macOS ARM64 the stack is reused more aggressively, the array was overwritten, and the hart ID lookup returned LIBRPMI_HSM_INVALID_HART_INDEX — causing RPMI_ERR_INVALID_PARAM (-3) instead of RPMI_SUCCESS.

Fixed by adding static to match that pattern, giving the array static storage duration for the lifetime of the program.

### Testing

All existing tests continue to pass.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,8 +39,8 @@ endif
 docs_dir=$(build_dir)/docs
 install_docs_dir=$(install_root_dir)/docs
 LIBRPMI_HEADER := $(src_dir)/include/librpmi.h
-LIBRPMI_IMPL_VERSION_MAJOR := $(strip $(shell sed -n 's/^#define[[:space:]]\+LIBRPMI_IMPL_VERSION_MAJOR[[:space:]]\+\([0-9][0-9]*\).*/\1/p' $(LIBRPMI_HEADER)))
-LIBRPMI_IMPL_VERSION_MINOR := $(strip $(shell sed -n 's/^#define[[:space:]]\+LIBRPMI_IMPL_VERSION_MINOR[[:space:]]\+\([0-9][0-9]*\).*/\1/p' $(LIBRPMI_HEADER)))
+LIBRPMI_IMPL_VERSION_MAJOR := $(strip $(shell grep 'define LIBRPMI_IMPL_VERSION_MAJOR' $(LIBRPMI_HEADER) | awk '{print $$NF}'))
+LIBRPMI_IMPL_VERSION_MINOR := $(strip $(shell grep 'define LIBRPMI_IMPL_VERSION_MINOR' $(LIBRPMI_HEADER) | awk '{print $$NF}'))
 ifeq ($(LIBRPMI_IMPL_VERSION_MAJOR),)
 $(error Failed to parse LIBRPMI_IMPL_VERSION_MAJOR from $(LIBRPMI_HEADER))
 endif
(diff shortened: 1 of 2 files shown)


icosta-sifive:
@pathakraul I found out that my previous [PR-106](https://github.com/riscv-software-src/librpmi/pull/106) I have introduced an error in the unit test, that is not visible in UBUNTU but only on MacOS. Plus i found out some issues on makefile that i have fixed. This highlighted the need of a light CI, i see that there is this open [PR-75](https://github.com/riscv-software-src/librpmi/pull/75) but is ion draft mode, so i took the opportunity to use what has been done in this PR + i have added more tests based on the CONTRIBUTING.md guidelins to start with. Once this one get merged i can propose a light github action PR

pathakraul:
```
GNU Make treats # as a comment character inside $(shell ...), silently truncating #define in the sed pattern and leaving $(strip unterminated. A secondary issue: \) in the sed capture group was counted by Make as a bare ) closing $(shell prematurely.
Fixed by replacing sed with grep | awk — no # in the Make expression.
```

On Ubuntu how is this an issue??

And if this is only on MacOS then what is the make version on that

Makefile
```diff
@@ -190,10 +187,18 @@ compile_elf = $(CMD_PREFIX)mkdir -p `dirname $(1)`; \
 compile_ar = $(CMD_PREFIX)mkdir -p `dirname $(1)`; \
 	     echo " AR        $(subst $(build_dir)/,,$(1))"; \
 	     $(AR) $(ARFLAGS) $(1) $(2)
+ifeq ($(shell uname -s),Darwin)
+compile_so = $(CMD_PREFIX)mkdir -p `dirname $(1)`; \
+	     echo " SO        $(subst $(build_dir)/,,$(1))"; \
+	     $(CC) -shared -Wl,-install_name,librpmi.so.$(LIBRPMI_SOVERSION) \
+	       -Wl,-undefined,dynamic_lookup \
```
saini-ranbirs: See if just updating the EXTRA_LDFLAGS earlier is sufficient, so that the existing **compile_so** part remains common.

icosta-sifive:
> ```
> GNU Make treats # as a comment character inside $(shell ...), silently truncating #define in the sed pattern and leaving $(strip unterminated. A secondary issue: \) in the sed capture group was counted by Make as a bare ) closing $(shell prematurely.
> Fixed by replacing sed with grep | awk — no # in the Make expression.
> ```
> 
> On Ubuntu how is this an issue??
> 
> And if this is only on MacOS then what is the make version on that

The # comment issue affects GNU Make < 4.3. Ubuntu 24.04 ships Make 4.3 which fixed this behaviour. macOS runners use Xcode's bundled Make 3.81 where # inside $(shell ...) is still silently treated as a comment. The grep | awk fix is correct and portable across all Make versions.

Makefile
icosta-sifive: @saini-ranbirs Thanks for the feedback. The platform-specific flags now live in EXTRA_LDFLAGS before the single common compile_so definition as suggested.

saini-ranbirs (approved):
LGTM

PR #113, opened 2026-08-18
build: add GitHub Actions CI workflows

icosta-sifive:
## Summary

- Add `unit.yml`: runs `make check` on ubuntu-24.04 and macos-15 on push/PR to main.
  Runners are pinned (not `*-latest`) to avoid silent breakage when GitHub shifts a runner OS.
- Add `pr-checks.yml`: three PR-only jobs:
  - **checkpatch** — runs `checkpatch.pl --no-tree` over all non-merge commits; covers DCO
    (Signed-off-by presence/correctness) and kernel-style code hygiene. Replaces a hand-rolled
    grep-based DCO check. `FILE_PATH_CHANGES` is suppressed (kernel MAINTAINERS convention,
    not applicable here).
  - **commit-format** — validates that commit subjects start with one of the project's required
    prefixes: `lib:` `include:` `test:` `docs:` `debian:` `build:`. checkpatch.pl has no
    concept of project-specific prefixes so this check is kept separate.
  - **docs** — runs `make docs` only when `docs/**` or `include/**` files change, gated via
    `dorny/paths-filter`.
 - Add `htmldocs` Makefile target: runs doxygen without the LaTeX/PDF step. The existing `docs`
  target depends on `htmldocs` so the full PDF build is preserved for local use. CI uses
  `htmldocs` to avoid requiring `pdflatex` on the runner.

## Notes

- The project uses Linux kernel commit style, not Conventional Commits — off-the-shelf
  `commitlint` would reject valid commits, hence the custom prefix check.
- checkpatch.pl is downloaded fresh from `torvalds/linux` at CI runtime (no vendored copy).

## Test plan

- [x] Push a commit missing `Signed-off-by:` — checkpatch job should fail
image

- [x] Push a commit with a subject that lacks a required prefix — commit-format job should fail
image

- [x] Push a docs change — docs job should run and pass
image

- [x] Push a non-docs change — docs job should be skipped
image

- [x] All jobs green on a well-formed PR
image

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -1,0 +1,69 @@
+name: PR Checks
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+    branches:
+      - main
+
+jobs:
+  checkpatch:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Install checkpatch
+        run: |
+          wget -q https://raw.githubusercontent.com/torvalds/linux/master/scripts/checkpatch.pl
+          wget -q https://raw.githubusercontent.com/torvalds/linux/master/scripts/spelling.txt
+          wget -q https://raw.githubusercontent.com/torvalds/linux/master/scripts/const_structs.checkpatch
+          chmod +x checkpatch.pl
+      - name: Run checkpatch
+        run: |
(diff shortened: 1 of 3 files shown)


icosta-sifive:
@pathakraul in order to make this work, some settings need to be done on the repo:

1. Go to https://github.com/icosta-sifive/librpmi/settings/branches
2. Click Add branch protection rule
3. Branch name pattern: main
4. Check Require status checks to pass before merging
5. Search for and add the job names:
- unit-test
- checkpatch
- commit-format
- docs
6. Optionally check Require branches to be up to date before merging
7. Click Save changes

PR #114, opened 2026-08-20
lib,test: align logging service and add coverage

slingappa:
Summary:
- align LOGGING service 0x02 with LOG_DATA request format
- allow LOGGING from S-mode and keep notifications unsupported
- add test_srvgrp_logging coverage

Testing:
- make O=/tmp/librpmi-build-final-cross LIBRPMI_TEST=y
- make O=/tmp/librpmi-build-final-host LIBRPMI_TEST=y CROSS_COMPILE=
- /tmp/librpmi-build-final-host/test/test_srvgrp_logging.elf

### DIFF
--- a/include/librpmi.h
+++ b/include/librpmi.h
@@ -198,7 +198,7 @@ enum rpmi_servicegroup_id {
 	RPMI_SRVGRP_MANAGEMENT_MODE	= 0x000B,
 	RPMI_SRVGRP_RAS_AGENT		= 0x000C,
 	RPMI_SRVGRP_REQUEST_FORWARD	= 0x000D,
-	RPMI_SRVGRP_LOGGING			= 0x000E,
+	RPMI_SRVGRP_LOGGING		= 0x000E,
 	RPMI_SRVGRP_ID_MAX_COUNT,
 
 	/* Reserved range for service groups */
@@ -373,7 +373,7 @@ enum rpmi_mm_service_id {
 /** RPMI LOGGING ServiceGroup Service IDs */
 enum rpmi_logging_service_id {
 	RPMI_LOGGING_SRV_ENABLE_NOTIFICATION = 0x01,
-	RPMI_LOGGING_SRV_SET_CONFIG = 0x02,
+	RPMI_LOGGING_SRV_LOG_DATA = 0x02,
 	RPMI_LOGGING_SRV_ID_MAX,
 };
 
@@ -2132,9 +2132,9 @@ enum rpmi_error rpmi_mm_service_register(struct rpmi_service_group *group,
 
 /** Platform specific logging operations */
 struct rpmi_logging_platform_ops {
-	enum rpmi_error (*do_set_state)(void *priv, rpmi_uint32_t log_type,
-					rpmi_uint32_t datalen_bytes,
-					const void *data);
+	enum rpmi_error (*log_data)(void *priv, rpmi_uint32_t log_type,
+				    rpmi_uint32_t num_words,
+				    const rpmi_uint32_t *words);
 };
 
 /**
--- a/lib/rpmi_service_group_logging.c
+++ b/lib/rpmi_service_group_logging.c
@@ -20,61 +20,82 @@ struct rpmi_logging_group {
 };
 
 static enum rpmi_error
-rpmi_logging_set_state(struct rpmi_logging_group *sglogging,
-		      rpmi_uint32_t log_type, rpmi_uint32_t datalen_bytes,
-		      const void *data)
+rpmi_logging_log_data(struct rpmi_logging_group *sglogging,
+		      rpmi_uint32_t log_type, rpmi_uint32_t num_words,
+		      const rpmi_uint32_t *words)
 {
 	enum rpmi_error ret;
 
-	ret = sglogging->ops->do_set_state(sglogging->ops_priv, log_type,
-					   datalen_bytes, data);
+	ret = sglogging->ops->log_data(sglogging->ops_priv, log_type,
+				       num_words, words);
 
 	return ret;
 }
 
 static enum rpmi_error
-rpmi_logging_sg_set_config(struct rpmi_service_group *group,
-			   struct rpmi_service *service,
-			   struct rpmi_transport *trans,
-			   rpmi_uint16_t request_datalen,
-			   const rpmi_uint8_t *request_data,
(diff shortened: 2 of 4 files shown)


include/librpmi.h
saini-ranbirs: Not directly related, but may be a chance to align LOGGING with all others

lib/rpmi_service_group_logging.c
```diff
@@ -20,49 +20,69 @@ struct rpmi_logging_group {
...
+	remaining_words = (request_datalen - (2 * sizeof(*req))) / sizeof(*req);
 
-	/* change logging config synchronously */
-	status = rpmi_logging_set_state(logginggrp, log_type, datalen_bytes,
-					data);
+	if (num_words != remaining_words) {
+		resp[0] = rpmi_to_xe32(trans->is_be,
+				       (rpmi_uint32_t)RPMI_ERR_INVALID_PARAM);
+		*response_datalen = sizeof(*resp);
+		return RPMI_SUCCESS;
```
saini-ranbirs: See if to just update status and do a common handling at the end using done: label

lib/rpmi_service_group_logging.c
```diff
@@ -20,49 +20,69 @@ struct rpmi_logging_group {
...
+	}
+
+	if (num_words) {
+		words = rpmi_env_zalloc(num_words * sizeof(*words));
+		if (!words) {
+			resp[0] = rpmi_to_xe32(trans->is_be,
+					       (rpmi_uint32_t)RPMI_ERR_FAILED);
+			*response_datalen = sizeof(*resp);
+			return RPMI_SUCCESS;
```
saini-ranbirs: See if to just update status and do a common handling at the end using done: label

lib/rpmi_service_group_logging.c
```diff
@@ -20,49 +20,69 @@ struct rpmi_logging_group {
...
+		}
+
+		for (i = 0; i < num_words; i++)
+			words[i] = rpmi_to_xe32(trans->is_be, req[2 + i]);
+	}
+
+	status = rpmi_logging_log_data(logginggrp, log_type, num_words, words);
+	if (words)
+		rpmi_env_free(words);
```
saini-ranbirs, line 80: Better give a blank line after statement under if ends. Also, cross check if **words** can really be NULL at this point.
Label can be added following that before next statement.

lib/rpmi_service_group_logging.c
```diff
@@ -20,49 +20,69 @@ struct rpmi_logging_group {
...
 	resp[0] = rpmi_to_xe32(trans->is_be, (rpmi_uint32_t)status);
```
saini-ranbirs, line 81: This blank line can be removed

saini-ranbirs (changes requested):

include/librpmi.h
slingappa: Aligned `RPMI_SRVGRP_LOGGING` with the surrounding service-group enum entries. This thread is marked outdated now, but the current code includes the cleanup.

lib/rpmi_service_group_logging.c
slingappa: Updated this path to set `status` and jump to the shared `done:` block. The response status and response length are now written in one place.

lib/rpmi_service_group_logging.c
slingappa: Done. The allocation failure path now sets `RPMI_ERR_FAILED` and uses the common `done:` response handling.

lib/rpmi_service_group_logging.c
slingappa: Added the `done:` label and spacing around the cleanup path. I kept the `if (words)` guard because `words` can be NULL for zero-word LOG_DATA requests and for error paths before allocation.

lib/rpmi_service_group_logging.c
slingappa: Removed the extra blank line by folding the response handling into the shared `done:` block.

lib/rpmi_service_group_logging.c
```diff
@@ -20,61 +20,82 @@ struct rpmi_logging_group {
...
+			goto done;
+		}
+
+		for (i = 0; i < num_words; i++)
+			words[i] = rpmi_to_xe32(trans->is_be, req[2 + i]);
+	}
 
+	status = rpmi_logging_log_data(logginggrp, log_type, num_words, words);
+
+done:
```
saini-ranbirs: Will there ever be a scenario of reaching done from goto with words as non-null?

If not, the label should be moved before line 83.

lib/rpmi_service_group_logging.c
```diff
@@ -20,61 +20,82 @@ struct rpmi_logging_group {
...
-	rpmi_uint32_t datalen_bytes;
-
-	if (request_datalen < sizeof(rpmi_uint32_t)) {
-		resp[0] = rpmi_to_xe32(trans->is_be,
-				       (rpmi_uint32_t)RPMI_ERR_INVALID_PARAM);
-		*response_datalen = sizeof(*resp);
-		return RPMI_SUCCESS;
+	const rpmi_uint32_t *req = (const rpmi_uint32_t *)request_data;
+	rpmi_uint32_t log_type, num_words, remaining_words, i;
+	rpmi_uint32_t *words = NULL;
```
saini-ranbirs, line 48: See if to follow reverse xmas tree for variable declaration part - https://www.kernel.org/doc/html/v6.2/process/maintainer-netdev.html

lib/rpmi_service_group_logging.c
slingappa: Moved the `done:` label below the normal-path `words` cleanup and folded it into `lib: align logging service with LOG_DATA` (`62967ca`). The `goto done` paths happen before a successful `words` allocation, or on allocation failure where `words` is still NULL, so they now jump directly to response serialization. The successful callback path still frees `words` before reaching `done:`.

lib/rpmi_service_group_logging.c
slingappa: Reordered the local declarations in `rpmi_logging_sg_log_data()` into reverse-Xmas-tree order and folded it into `lib: align logging service with LOG_DATA` (`62967ca`).

saini-ranbirs (approved):

PR #115, opened 2026-08-24
lib: cppc: add autonomous (CPPC2) mode support

icosta-sifive:
This PR implements autonomous (CPPC2) mode support for the CPPC service group as per the RPMI specification.

## Overview
Adds full support for autonomous mode alongside the existing passive (normal) mode. In autonomous mode the application processor provides minimum and maximum performance bounds via the fastchannel, and the platform firmware selects the actual operating point within those bounds. This is in contrast to passive mode where the AP writes an explicit desired performance level.

Also fixes a spec compliance issue in CPPC_PROBE_REG: a valid but mode-unsupported register must return SUCCESS with REG_LENGTH=0, not RPMI_ERR_NOTSUPP.

### Library Implementation (lib/rpmi_service_group_cppc.c)
- Added cppc_update_perf_auto() platform callback for autonomous mode performance bound updates
- Unlocked RPMI_CPPC_AUTO_MODE in the create function and store the mode at runtime
- Implemented probe, read, and write for the 10 previously unimplemented registers, gating autonomous-only registers on mode
- Added autonomous mode event loop reading min_perf/max_perf from the fastchannel active union
GET_FAST_CHANNEL_REGION now encodes FLAGS[4:3] = 0b01 for autonomous mode

### Public API (include/librpmi.h)

- Added cppc_update_perf_auto() callback to rpmi_cppc_platform_ops for autonomous mode performance bound updates

### Tests (test/test_srvgrp_cppc.c)

- Updated probe_not_supp_expdata to expect SUCCESS + REG_LENGTH=0 following the spec compliance fix
- Added test_cppc_update_perf_auto() stub and wired it into the shared test_cppc_ops
- Added scenario_cppc_auto with 10 test cases covering:
PROBE_REG for autonomous-only registers returns SUCCESS + 32 in autonomous mode
PROBE_REG for both-modes registers (GUARANTEED_PERF, COUNTER_WRAPAROUND_TIME) returns correct lengths
READ_REG for GUARANTEED_PERF and COUNTER_WRAPAROUND_TIME from the static regs struct
READ_REG and WRITE_REG for ENERGY_PERF_PREFERENCE via platform callbacks
WRITE_REG for MIN_PERF returns DENIED when fastchannel is active
GET_FAST_CHANNEL_REGION returns FLAGS = 0x08 in autonomous mode
- Updated main() to run both passive and autonomous scenarios

### Testing

All existing passive mode tests continue to pass

## Compliance
Implements autonomous (CPPC2) mode for the CPPC service group as specified in the RISC-V RPMI specification. .

### DIFF
--- a/include/librpmi.h
+++ b/include/librpmi.h
@@ -1498,17 +1498,26 @@ struct rpmi_cppc_platform_ops {
 					rpmi_uint32_t hart_index,
 					rpmi_uint64_t val);
 	/**
-	 * cppc update performance level for a hart
+	 * cppc update performance level for a hart (passive/normal mode)
 	 */
 	enum rpmi_error (*cppc_update_perf)(void *priv,
 					    rpmi_uint32_t hart_index,
 					    rpmi_uint32_t desired_perf);
+
 	/**
 	 * cppc get current frequency in hertz for a hart
 	 */
 	enum rpmi_error (*cppc_get_current_freq)(void *priv,
 						 rpmi_uint32_t hart_index,
 						 rpmi_uint64_t *current_freq_hz);
+
+	/**
+	 * cppc update performance bounds for a hart (autonomous mode)
+	 */
+	enum rpmi_error (*cppc_update_perf_auto)(void *priv,
+						 rpmi_uint32_t hart_index,
+						 rpmi_uint32_t min_perf,
+						 rpmi_uint32_t max_perf);
 };
 
 /**
--- a/lib/rpmi_service_group_cppc.c
+++ b/lib/rpmi_service_group_cppc.c
@@ -665,16 +665,10 @@ rpmi_cppc_sg_get_fast_channel_region(struct rpmi_service_group *group,
 	resp[6] = 0;
 	/* doorbell addr high */
 	resp[7] = 0;
-	/* doorbell set mask low */
+	/* doorbell write value */
 	resp[8] = 0;
-	/* doorbell set mask high */
-	resp[9] = 0;
-	/* doorbell preserve mask low */
-	resp[10] = 0;
-	/* doorbell preserve mask high */
-	resp[11] = 0;
 
-	resp_dlen = 12 * sizeof(*resp);
+	resp_dlen = 9 * sizeof(*resp);
 
 done:
 	resp[0] = rpmi_to_xe32(trans->is_be, (rpmi_uint32_t)status);
--- a/lib/rpmi_service_group_cppc.c
+++ b/lib/rpmi_service_group_cppc.c
@@ -58,6 +58,9 @@ struct rpmi_cppc_group {
 	/** CPPC mode of operation */
 	enum rpmi_cppc_mode cppc_mode;
 
+	/** current value of AutonomousSelectionEnable register */
+	rpmi_uint32_t autonomous_selection_enable;
+
 	/**
 	 * hsm context representing all harts managed
 	 * by CPPC service group
@@ -114,7 +117,7 @@ __cppc_hart_fc_perf_feedback_offset(struct rpmi_cppc_fastchan *fastchan_ctx,
 }
 
 /**
- * Get the desired_perf value from fastchannel of a hart
+ * Get the desired_perf value from fastchannel of a hart (passive mode)
  */
 static inline rpmi_uint32_t
 __cppc_get_fc_desired_perf(struct rpmi_cppc_group *cppcgrp,
@@ -125,6 +128,49 @@ __cppc_get_fc_desired_perf(struct rpmi_cppc_group *cppcgrp,
 	rpmi_uint64_t offset;
 
 	offset = __cppc_hart_fc_perf_request_offset(cppcgrp->fastchan_ctx, hart_index);
+
+	rc = rpmi_shmem_read(cppcgrp->fastchan_ctx->shmem, offset, &val,
+			     sizeof(rpmi_uint32_t));
+	if (rc)
+		return 0;
+
+	return val;
+}
+
+/**
+ * Get the min_perf value from fastchannel of a hart (autonomous mode)
+ */
+static inline rpmi_uint32_t
+__cppc_get_fc_min_perf(struct rpmi_cppc_group *cppcgrp, rpmi_uint32_t hart_index)
+{
+	int rc;
+	rpmi_uint32_t val;
+	rpmi_uint64_t offset;
+
+	/* active.min_perf is the first u32 of the fastchannel entry */
+	offset = __cppc_hart_fc_perf_request_offset(cppcgrp->fastchan_ctx, hart_index);
+
+	rc = rpmi_shmem_read(cppcgrp->fastchan_ctx->shmem, offset, &val,
+			     sizeof(rpmi_uint32_t));
+	if (rc)
+		return 0;
+
+	return val;
+}
+
+/**
+ * Get the max_perf value from fastchannel of a hart (autonomous mode)
+ */
+static inline rpmi_uint32_t
+__cppc_get_fc_max_perf(struct rpmi_cppc_group *cppcgrp, rpmi_uint32_t hart_index)
+{
+	int rc;
+	rpmi_uint32_t val;
+	rpmi_uint64_t offset;
+
+	/* active.max_perf is the second u32 of the fastchannel entry */
+	offset = __cppc_hart_fc_perf_request_offset(cppcgrp->fastchan_ctx, hart_index)
+		 + sizeof(rpmi_uint32_t);
 
 	rc = rpmi_shmem_read(cppcgrp->fastchan_ctx->shmem, offset, &val,
 			     sizeof(rpmi_uint32_t));
@@ -184,17 +230,36 @@ static enum rpmi_error __rpmi_cppc_probe_reg(struct rpmi_cppc_group *cppcgrp,
 		status = RPMI_SUCCESS;
 		len = BITS(sizeof(rpmi_uint64_t));
 		break;
-	/** Not implemented registers (will be in future) */
-	case RPMI_CPPC_MAX_PERF:
-	case RPMI_CPPC_MIN_PERF:
+	/* Available in both modes */
 	case RPMI_CPPC_GUARANTEED_PERF:
-	case RPMI_CPPC_TIME_WINDOW:
-	case RPMI_CPPC_PERF_REDUCTION_TOLERANCE:
-	case RPMI_CPPC_AUTONOMOUS_SELECTION_ENABLE:
 	case RPMI_CPPC_CPPC_ENABLE:
+		status = RPMI_SUCCESS;
+		len = BITS(sizeof(rpmi_uint32_t));
+		break;
 	case RPMI_CPPC_COUNTER_WRAPAROUND_TIME:
+		status = RPMI_SUCCESS;
+		len = BITS(sizeof(rpmi_uint64_t));
+		break;
+	/* Available in both modes */
+	case RPMI_CPPC_MIN_PERF:
+	case RPMI_CPPC_MAX_PERF:
+		status = RPMI_SUCCESS;
+		len = BITS(sizeof(rpmi_uint32_t));
+		break;
+	/* Autonomous (CPPC2) mode only */
+	case RPMI_CPPC_PERF_REDUCTION_TOLERANCE:
+	case RPMI_CPPC_TIME_WINDOW:
+	case RPMI_CPPC_AUTONOMOUS_SELECTION_ENABLE:
 	case RPMI_CPPC_AUTONOMOUS_ACTIVITY_WINDOW:
 	case RPMI_CPPC_ENERGY_PERF_PREFERENCE:
+		if (cppcgrp->cppc_mode != RPMI_CPPC_AUTO_MODE) {
+			status = RPMI_ERR_NOTSUPP;
+			len = 0;
+			break;
+		}
+		status = RPMI_SUCCESS;
+		len = BITS(sizeof(rpmi_uint32_t));
+		break;
 	default:
 		DPRINTF("%s: CPPC reg-%u not implemented\n", __func__, reg_id);
 		status = RPMI_ERR_NOTSUPP;
@@ -264,6 +329,49 @@ static enum rpmi_error __rpmi_cppc_read_reg(struct rpmi_cppc_group *cppcgrp,
 	case RPMI_CPPC_TRANSITION_LATENCY:
 		val = cppcgrp->regs->transition_latency;
 		break;
+	/* Available in both modes, static from regs struct */
+	case RPMI_CPPC_GUARANTEED_PERF:
+		val = cppcgrp->regs->guaranteed_perf;
+		break;
+	case RPMI_CPPC_COUNTER_WRAPAROUND_TIME:
+		val = cppcgrp->regs->counter_wraparound_time;
+		break;
+	/* Available in both modes, delegated to platform */
+	case RPMI_CPPC_CPPC_ENABLE:
+		status = cppcgrp->ops->cppc_get_reg(cppcgrp->ops_priv,
+						    RPMI_CPPC_CPPC_ENABLE,
+						    hart_index, &val);
+		break;
+	/* Available in both modes: fastchannel in autonomous mode if present */
+	case RPMI_CPPC_MIN_PERF:
+		if (cppcgrp->cppc_mode == RPMI_CPPC_AUTO_MODE &&
+		    cppcgrp->autonomous_selection_enable &&
+		    cppcgrp->fastchan_ctx)
+			val = __cppc_get_fc_min_perf(cppcgrp, hart_index);
+		else
+			status = cppcgrp->ops->cppc_get_reg(cppcgrp->ops_priv,
+							    RPMI_CPPC_MIN_PERF,
+							    hart_index, &val);
+		break;
+	case RPMI_CPPC_MAX_PERF:
+		if (cppcgrp->cppc_mode == RPMI_CPPC_AUTO_MODE &&
+		    cppcgrp->autonomous_selection_enable &&
+		    cppcgrp->fastchan_ctx)
+			val = __cppc_get_fc_max_perf(cppcgrp, hart_index);
+		else
+			status = cppcgrp->ops->cppc_get_reg(cppcgrp->ops_priv,
+							    RPMI_CPPC_MAX_PERF,
+							    hart_index, &val);
+		break;
+	/* Autonomous mode only, delegated to platform */
+	case RPMI_CPPC_PERF_REDUCTION_TOLERANCE:
+	case RPMI_CPPC_TIME_WINDOW:
+	case RPMI_CPPC_AUTONOMOUS_SELECTION_ENABLE:
+	case RPMI_CPPC_AUTONOMOUS_ACTIVITY_WINDOW:
+	case RPMI_CPPC_ENERGY_PERF_PREFERENCE:
+		status = cppcgrp->ops->cppc_get_reg(cppcgrp->ops_priv,
+						    reg_id, hart_index, &val);
+		break;
 	default:
 		/* No read permission */
 		status = RPMI_ERR_DENIED;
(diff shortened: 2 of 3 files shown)


include/librpmi.h
```diff
@@ -1498,11 +1498,20 @@ struct rpmi_cppc_platform_ops {
...
+	 * cppc update performance level for a hart (passive/normal mode)
 	 */
 	enum rpmi_error (*cppc_update_perf)(void *priv,
 					    rpmi_uint32_t hart_index,
 					    rpmi_uint32_t desired_perf);
+
+	/**
+	 * cppc update performance bounds for a hart (autonomous mode)
+	 */
+	enum rpmi_error (*cppc_update_perf_auto)(void *priv,
```
slingappa: Adding `cppc_update_perf_auto` in the middle of `struct rpmi_cppc_platform_ops` changes the layout of the public callback table. Existing binaries built with the old header place `cppc_get_current_freq` at this offset; with the new library, `rpmi_cppc_process_events()` can call that old pointer as `cppc_update_perf_auto` and then read `cppc_get_current_freq` past the old struct. This can break even passive-mode users when fast-channel events are processed, while the shared library `SOVERSION` remains `0`. Please keep the existing fields at stable offsets, for example by appending the new callback at the end, and reject autonomous mode in `rpmi_service_group_cppc_create()` when the new callback is missing.

test/test_srvgrp_cppc.c
```diff
@@ -421,8 +442,308 @@ static struct rpmi_test_scenario scenario_cppc_default = {
...
+	.reference_perf = TEST_CPPC_NOMINAL_PERF,
+	.lowest_freq = 1000000000U,
+	.nominal_freq = 2000000000U,
+	.transition_latency = 10,
+	.counter_wraparound_time = ((rpmi_uint64_t)TEST_CPPC_WRAPAROUND_HI << 32) |
+				   TEST_CPPC_WRAPAROUND_LO,
+};
+
+/* PROBE: autonomous-mode-only register → SUCCESS + 32 */
+static rpmi_uint32_t probe_min_perf_reqdata[] = {
```
slingappa, line 539: The new autonomous CPPC request payloads keep using `{ HART_ID, REG_ID }`, but the latest root RPMI spec includes CPPC_PROBE_REG, CPPC_READ_REG, and CPPC_WRITE_REG tables with word 0 as `REG_ID` and word 1 as `HART_ID`. The tests pass because the implementation parses the same swapped order, but a spec-compliant caller sending `{ RPMI_CPPC_MIN_PERF, hart0 }` would be interpreted as `hart_id = RPMI_CPPC_MIN_PERF` and fail with `RPMI_ERR_INVALID_PARAM`. Please update the CPPC service parsing and the tests to the specified wire order.

lib/rpmi_service_group_cppc.c
```diff
@@ -351,6 +471,14 @@ rpmi_cppc_sg_probe_reg(struct rpmi_service_group *group,
 	}
 
 	status = __rpmi_cppc_probe_reg(cppcgrp, cppc_reg_id, &reg_len);
+	/*
+	 * Per RPMI spec: a valid but unsupported register returns SUCCESS
+	 * with REG_LENGTH=0. The invalid-register check above guards against
+	 * truly out-of-range IDs, so any NOTSUPP here means "valid but not
+	 * implemented in this mode".
+	 */
+	if (status == RPMI_ERR_NOTSUPP)
```
slingappa: This remaps `RPMI_ERR_NOTSUPP` to `RPMI_SUCCESS` for CPPC_PROBE_REG, but the latest CPPC_PROBE_REG response table still lists `RPMI_ERR_NOT_SUPPORTED` for an unsupported `REG_ID`; the `REG_LENGTH` field is what becomes `0`. With this change, passive mode reports auto-only registers as successfully implemented with length 0, which makes it hard for callers to distinguish unsupported registers from implemented zero-length data. Please preserve the `NOTSUPP` status and keep `REG_LENGTH = 0`, unless there is a newer normative text that intentionally changed the status semantics.

test/test_srvgrp_cppc.c
```diff
@@ -421,8 +442,308 @@ static struct rpmi_test_scenario scenario_cppc_default = {
...
+	exp[4] = TEST_CPPC_FASTCHAN_SIZE;
+	exp[5] = 0;
+	exp[6] = 0;
+	exp[7] = 0;
+	exp[8] = 0;
+	exp[9] = 0;
+	exp[10] = 0;
+	exp[11] = 0;
+
+	return 12 * sizeof(*exp);
```
slingappa: The new autonomous GET_FAST_CHANNEL_REGION test expects 12 response words, including doorbell set/preserve mask fields, but the current root spec's CPPC_GET_FAST_CHANNEL_REGION response ends at word 8 (`DB_WRITE_VALUE`). Since this PR is adding CPPC2 coverage against the current fast-channel flags, please update the implementation/tests to the current 9-word response shape or document why librpmi intentionally keeps compatibility with the older layout.

lib/rpmi_service_group_cppc.c
```diff
@@ -680,34 +808,51 @@ static struct rpmi_service rpmi_cppc_services[RPMI_CPPC_SRV_ID_MAX] = {
...
+
+			if (hart_perf_request->active.min_perf != min_perf ||
+			    hart_perf_request->active.max_perf != max_perf) {
+				hart_perf_request->active.min_perf = min_perf;
+				hart_perf_request->active.max_perf = max_perf;
+				status = cppcgrp->ops->cppc_update_perf_auto(
+							cppcgrp->ops_priv,
+							hart_idx,
+							min_perf, max_perf);
+				cppcgrp->ops->cppc_get_current_freq(cppcgrp->ops_priv,
```
slingappa: The autonomous event path ignores the return value from `cppc_get_current_freq()` and then writes `current_freq` to the feedback fast-channel. If the platform callback fails without filling the output, this writes an uninitialized stack value and hides the failure by overwriting `status` with the fast-channel write status. Please initialize `current_freq`, check the callback return, and decide whether to skip feedback update or preserve the last known value on error.

test/test_srvgrp_cppc.c
```diff
@@ -421,8 +442,308 @@ static struct rpmi_test_scenario scenario_cppc_default = {
...
+	if (!grp) {
+		printf("failed to create rpmi cppc auto service group");
+		return RPMI_ERR_FAILED;
+	}
+
+	rpmi_context_add_group(scene->cntx, grp);
+	return 0;
+}
+
+static struct rpmi_test_scenario scenario_cppc_auto = {
```
slingappa, line 708: The added autonomous scenario covers probe/read/write services and fast-channel flags, but it does not write new min/max values into the performance-request fast-channel and verify that `rpmi_context_process_all_events()` calls `cppc_update_perf_auto()` and updates the feedback channel. That leaves the new callback and event path effectively untested; please add a focused case that writes min/max to `test_auto_fastchan_mem`, runs event processing, and validates callback arguments plus feedback contents.

lib/rpmi_service_group_cppc.c
```diff
@@ -680,34 +808,51 @@ static struct rpmi_service rpmi_cppc_services[RPMI_CPPC_SRV_ID_MAX] = {
...
+			}
+		} else {
+			min_perf = __cppc_get_fc_min_perf(cppcgrp, hart_idx);
+			max_perf = __cppc_get_fc_max_perf(cppcgrp, hart_idx);
+
+			if (hart_perf_request->active.min_perf != min_perf ||
+			    hart_perf_request->active.max_perf != max_perf) {
+				hart_perf_request->active.min_perf = min_perf;
+				hart_perf_request->active.max_perf = max_perf;
+				status = cppcgrp->ops->cppc_update_perf_auto(
```
slingappa: `checkpatch.pl --no-tree --strict` reports five style checks on this diff: the `cppc_update_perf_auto(` call is split with a line ending in `(`, and the new arrays near `test/test_srvgrp_cppc.c:492`, `:500`, `:508`, and `:516` need blank lines after declarations. Please clean these up and rerun checkpatch.

slingappa (commented):
Reviewed against the latest RPMI spec root (`riscv-rpmi.adoc`, which includes `src/srvgrp-cppc.adoc`). Local validation: host `LIBRPMI_TEST=y` build and `make check` passed, direct CPPC test passed, RISC-V cross build with tests passed. I could not execute the RISC-V ELFs under `qemu-riscv64` because it is not available in PATH. Checkpatch reports 0 errors, 0 warnings, and 5 style checks.

include/librpmi.h
icosta-sifive: Good catch, thanks.

Moved cppc_update_perf_auto to the end of struct rpmi_cppc_platform_ops so existing fields keep their stable offsets — cppc_get_current_freq is back at its original position.
Added an early return in rpmi_service_group_cppc_create() that rejects RPMI_CPPC_AUTO_MODE when ops->cppc_update_perf_auto is NULL, so passive-mode callers are unaffected.

test/test_srvgrp_cppc.c
icosta-sifive: Good, All three handlers (rpmi_cppc_sg_probe_reg, rpmi_cppc_sg_read_reg, rpmi_cppc_sg_write_reg) now parse reqdata[0] as REG_ID and reqdata[1] as HART_ID, matching the spec wire order. DATA_LO/DATA_HI at [2]/[3] in WRITE_REG are unchanged.
All 14 reqdata arrays in the test file are updated to { REG_ID, HART_ID, ... } accordingly.

lib/rpmi_service_group_cppc.c
icosta-sifive: You're right, the normative table requires RPMI_ERR_NOT_SUPPORTED with REG_LENGTH = 0 — I had misread an earlier draft. Fixed: the NOTSUPP→SUCCESS remap is removed. The handler already writes reg_len (which stays 0) and a two-word response length before done:, so NOTSUPP naturally falls through with the correct wire encoding. The test expectation is updated to match.

test/test_srvgrp_cppc.c
icosta-sifive: Confirmed, the set/preserve mask fields are not in the current spec. Fixed: the handler now returns a 9-word response (STATUS through DB_WRITE_VALUE at resp[8]), and both test builder functions (init_fastchan_region_expected and init_auto_fastchan_region_expected) are updated to match.

lib/rpmi_service_group_cppc.c
icosta-sifive: Fixed in both the autonomous and passive paths.

test/test_srvgrp_cppc.c
icosta-sifive: Added a focused event-path test

lib/rpmi_service_group_cppc.c
```diff
@@ -264,6 +322,45 @@ static enum rpmi_error __rpmi_cppc_read_reg(struct rpmi_cppc_group *cppcgrp,
...
+	case RPMI_CPPC_MIN_PERF:
+		if (cppcgrp->fastchan_ctx)
+			val = __cppc_get_fc_min_perf(cppcgrp, hart_index);
+		else
+			status = cppcgrp->ops->cppc_get_reg(cppcgrp->ops_priv,
+							    RPMI_CPPC_MIN_PERF,
+							    hart_index, &val);
+		break;
+	case RPMI_CPPC_MAX_PERF:
+		if (cppcgrp->fastchan_ctx)
```
vlsunil: A check for autonomous mode is required as well, right?

lib/rpmi_service_group_cppc.c
```diff
@@ -296,13 +393,36 @@ static enum rpmi_error __rpmi_cppc_write_reg(struct rpmi_cppc_group *cppcgrp,
 						    hart_index,
 						    reg_val);
 		break;
+	/* Autonomous mode only: fastchannel if present, else platform */
+	case RPMI_CPPC_MIN_PERF:
+	case RPMI_CPPC_MAX_PERF:
+		if (cppcgrp->fastchan_ctx) {
+			status = RPMI_ERR_DENIED;
```
vlsunil: Why?

lib/rpmi_service_group_cppc.c
```diff
@@ -330,10 +450,10 @@ rpmi_cppc_sg_probe_reg(struct rpmi_service_group *group,
 	struct rpmi_cppc_group *cppcgrp = group->priv;
 	rpmi_uint32_t *resp = (void *)response_data;
 
-	hart_id = rpmi_to_xe32(trans->is_be,
-			((const rpmi_uint32_t *)request_data)[0]);
 	cppc_reg_id = rpmi_to_xe32(trans->is_be,
-			    ((const rpmi_uint32_t *)request_data)[1]);
+			    ((const rpmi_uint32_t *)request_data)[0]);
+	hart_id = rpmi_to_xe32(trans->is_be,
+			       ((const rpmi_uint32_t *)request_data)[1]);
```
vlsunil, line 472: Better to make this separate commit as a fix?

lib/rpmi_service_group_cppc.c
```diff
@@ -184,17 +227,32 @@ static enum rpmi_error __rpmi_cppc_probe_reg(struct rpmi_cppc_group *cppcgrp,
...
+		break;
+	/* Autonomous (CPPC2) mode only */
+	case RPMI_CPPC_MIN_PERF:
+	case RPMI_CPPC_MAX_PERF:
+	case RPMI_CPPC_PERF_REDUCTION_TOLERANCE:
+	case RPMI_CPPC_TIME_WINDOW:
+	case RPMI_CPPC_AUTONOMOUS_SELECTION_ENABLE:
 	case RPMI_CPPC_AUTONOMOUS_ACTIVITY_WINDOW:
 	case RPMI_CPPC_ENERGY_PERF_PREFERENCE:
+		if (cppcgrp->cppc_mode != RPMI_CPPC_AUTO_MODE) {
```
vlsunil, line 255: I don't see ACPI spec suggesting to implement MIN and MAX PERF registers only in AUTO mode. Let me know if I am missing something.

lib/rpmi_service_group_cppc.c
```diff
@@ -529,16 +649,10 @@ rpmi_cppc_sg_get_fast_channel_region(struct rpmi_service_group *group,
 	resp[6] = 0;
 	/* doorbell addr high */
 	resp[7] = 0;
-	/* doorbell set mask low */
+	/* doorbell write value */
```
vlsunil, line 668: IMO, this should be a separate fix commit.

lib/rpmi_service_group_cppc.c
```diff
@@ -296,13 +393,36 @@ static enum rpmi_error __rpmi_cppc_write_reg(struct rpmi_cppc_group *cppcgrp,
...
 						    reg_val);
 		break;
+	/* Autonomous mode only: fastchannel if present, else platform */
+	case RPMI_CPPC_MIN_PERF:
+	case RPMI_CPPC_MAX_PERF:
+		if (cppcgrp->fastchan_ctx) {
+			status = RPMI_ERR_DENIED;
+			break;
+		}
+		status = cppcgrp->ops->cppc_set_reg(cppcgrp->ops_priv,
```
vlsunil, line 415: Should there be a check that  MIN and MAX value in the range [Lowest Performance, Highest Performance].

lib/rpmi_service_group_cppc.c
icosta-sifive: good catch — fixed in both __rpmi_cppc_read_reg and __rpmi_cppc_write_reg. MIN_PERF and MAX_PERF now return RPMI_ERR_NOTSUPP immediately when cppc_mode != RPMI_CPPC_AUTO_MODE, consistent with the existing check in __rpmi_cppc_probe_reg.

lib/rpmi_service_group_cppc.c
icosta-sifive: You're right to question it. I Checked the spec: RPMI_ERR_DENIED in the WRITE_REG error table is scoped to read-only registers only. For the fast-channel case, the spec says writes to MinimumPerformanceRegister/MaximumPerformanceRegister via the message interface "may be ignored" when fast-channels are present — the AP is expected to use the fast-channel, but the platform firmware is not required to reject the write with an error. The DENIED here was a wrong analogy with DESIRED_PERF. Fixed: the write now goes through to cppc_set_reg regardless of fastchan, and the test expectation updated to RPMI_SUCCESS.

lib/rpmi_service_group_cppc.c
icosta-sifive: Your are right, Just double check. MinimumPerformanceRegister and MaximumPerformanceRegister are base ACPI CPPC registers available in both passive and autonomous modes — they bound the platform's operating range regardless of mode. Only PERF_REDUCTION_TOLERANCE, TIME_WINDOW, AUTONOMOUS_SELECTION_ENABLE, AUTONOMOUS_ACTIVITY_WINDOW, and ENERGY_PERF_PREFERENCE are CPPC2-only.

Fixed in probe, read, and write handlers: MIN_PERF and MAX_PERF are now always available. The read path still prefers the fastchannel in autonomous mode (since that's where the AP writes them), falling back to cppc_get_reg in passive mode or when no fastchan. The passive-mode "not supported" probe test is updated to use ENERGY_PERF_PREFERENCE instead.

lib/rpmi_service_group_cppc.c
icosta-sifive: Done — added range validation for MIN_PERF and MAX_PERF writes: values outside [LowestPerformance, HighestPerformance] now return RPMI_ERR_INVALID_PARAM before reaching cppc_set_reg.

lib/rpmi_service_group_cppc.c
```diff
@@ -264,6 +326,47 @@ static enum rpmi_error __rpmi_cppc_read_reg(struct rpmi_cppc_group *cppcgrp,
...
+		break;
+	/* Available in both modes, delegated to platform */
+	case RPMI_CPPC_CPPC_ENABLE:
+		status = cppcgrp->ops->cppc_get_reg(cppcgrp->ops_priv,
+						    RPMI_CPPC_CPPC_ENABLE,
+						    hart_index, &val);
+		break;
+	/* Available in both modes: fastchannel in autonomous mode if present */
+	case RPMI_CPPC_MIN_PERF:
+		if (cppcgrp->cppc_mode == RPMI_CPPC_AUTO_MODE &&
```
vlsunil, line 347: This cppc_mode can be treated as platform capability to support CPPC2 via _OSC. However, OSPM has a choice to enable it via "Autonomous Selection Enable" unless it is hardcoded to 1 by the platform. Hence, it should check whether "Autonomous Selection Enable" as well.

lib/rpmi_service_group_cppc.c
```diff
@@ -680,34 +807,54 @@ static struct rpmi_service rpmi_cppc_services[RPMI_CPPC_SRV_ID_MAX] = {
...
+			}
+		} else {
+			min_perf = __cppc_get_fc_min_perf(cppcgrp, hart_idx);
+			max_perf = __cppc_get_fc_max_perf(cppcgrp, hart_idx);
+
+			if (hart_perf_request->active.min_perf != min_perf ||
+			    hart_perf_request->active.max_perf != max_perf) {
+				hart_perf_request->active.min_perf = min_perf;
+				hart_perf_request->active.max_perf = max_perf;
+				status = cppcgrp->ops->cppc_update_perf_auto(cppcgrp->ops_priv,
```
vlsunil: What is the use of this API? Shouldn't the frequency update be done by the platform itself?

lib/rpmi_service_group_cppc.c
icosta-sifive: Good point. Added autonomous_selection_enable as a runtime field in rpmi_cppc_group, initialized from cppc_regs at create time and updated on WRITE_REG. The fastchannel path for MIN/MAX_PERF reads and the process_events autonomous loop now both gate on it in addition to cppc_mode.

lib/rpmi_service_group_cppc.c
icosta-sifive: @vlsunil cppc_update_perf_auto follows the same pattern as cppc_update_perf in passive mode — process_events reads the fastchannel and calls the platform ops callback. The difference is semantic: in passive mode the callback IS the performance update; in autonomous mode it's a notification that bounds changed so the platform can immediately reconsider its operating point. On a true hardware-autonomous platform this is redundant since the hardware picks up new bounds itself. We can make it optional (call only if non-NULL) so platforms that don't need it don't have to provide a stub — let me know.

vlsunil: Okay. Thanks!. Yes, better to add NULL check.

test/test_srvgrp_cppc.c
```diff
@@ -421,8 +518,442 @@ static struct rpmi_test_scenario scenario_cppc_default = {
...
+	exp[4] = TEST_CPPC_FASTCHAN_SIZE;
+	exp[5] = 0;
+	exp[6] = 0;
+	exp[7] = 0;
+	exp[8] = 0;
+	exp[9] = 0;
+	exp[10] = 0;
+	exp[11] = 0;
+
+	return 12 * sizeof(*exp);
```
vlsunil: Shouldn't this be 9 elements?

lib/rpmi_service_group_cppc.c
icosta-sifive: Done

test/test_srvgrp_cppc.c
icosta-sifive: That was already handled

vlsunil (approved):

PR #116, opened 2026-08-26
voltage: fix status reporting and linear range level packing

yeongjoshua:
This PR fixes two conformance bugs in the VOLTAGE service group (SERVICEGROUP_ID 0x0007) as specified in the RPMI specification, where the replies built by the library did not match the encoding the specification defines.

### Overview

The VOLTAGE service group lets a supervisor enumerate the voltage domains of a platform and read or change their levels. Two of its services built replies that a conforming client cannot use: a successful attributes query did not report its status, and the supported levels of a linear range domain were truncated to a third of their width. Both are in the reply framing, which the library owns, so a platform implementation cannot work around either of them.

### Changes

1. Library implementation (lib/rpmi_service_group_voltage.c) — fixes 2 of the 8 services defined for the group.

VOLT_GET_ATTRIBUTES (0x03) — rpmi_volt_get_attributes() fills in the capability, level count, transition latency and domain name of a successful reply, but writes resp[0] only on its error paths, unlike every other service in the group. On success the status word keeps whatever the previous response left in the buffer, so a valid query can report an error while carrying correct attributes. To reproduce: query an invalid domain, then a valid one — the second reply carries the RPMI_ERR_INVALID_PARAM of the first.

VOLT_GET_SUPPORTED_LEVELS (0x04) — the spec types VOLTAGE_LEVEL[] as uint32 or uint32[3], and says NUM_LEVELS/RETURNED/REMAINING count levels, each linear range counting as one. The reply was packed as one word per level regardless of format, so a linear domain returned only its voltage_min. Fixed by deriving the width of a level from the advertised format and applying it to the copy loop, response_datalen, and the message capacity (which counted words rather than levels).

2. Tests (test/test_srvgrp_voltage.c) — fixes the linear test domain, which was declared with a two-word level array and num_levels = ARRAY_SIZE(), so correct packing walked off the end of it into the neighbouring discrete table; counts and indexes its levels in tuples; validates a linear level against its range; updates the expected reply to the full tuple; and adds a regression test for the status bug.

### Testing

- [x]  make — builds successfully without warnings

- [x] make LIBRPMI_TEST=y — builds with tests successfully

- [x]  make LIBRPMI_TEST=y check — all tests pass, 134 total, 17 in the voltage service group

Each fix is covered by a test that fails without it.

Compliance — brings VOLT_GET_ATTRIBUTES and VOLT_GET_SUPPORTED_LEVELS into line with the VOLTAGE service group as specified in the RISC-V RPMI specification.

### DIFF
--- a/include/librpmi.h
+++ b/include/librpmi.h
@@ -1958,6 +1958,10 @@ enum rpmi_voltage_type {
 	(((__format) & RPMI_VOLT_CAPABILITY_FORMAT_MASK) <<		\
 	 RPMI_VOLT_CAPABILITY_FORMAT_POS)
 
+#define RPMI_VOLT_CAPABILITY_GET_FORMAT(__capability)			\
+	(((__capability) >> RPMI_VOLT_CAPABILITY_FORMAT_POS) &		\
+	 RPMI_VOLT_CAPABILITY_FORMAT_MASK)
+
 /** voltage domain capabilities */
 enum rpmi_voltage_capability {
 	RPMI_VOLT_CAPABILITY_INVALID		= -1,
@@ -1990,6 +1994,18 @@ struct rpmi_voltage_linear_range {
  * This structure represents the static
  * voltage domain data which platform has to maintain
  * and pass to create the voltage service group.
+ *
+ * The layout of the voltage level array depends on voltage_type. The
+ * num_levels field below, the MAX_NUM_LEVELS returned by VOLT_GET_ATTRIBUTES
+ * and the RETURNED and REMAINING counts of VOLT_GET_SUPPORTED_LEVELS all
+ * count voltage levels, never words:
+ *
+ * RPMI_VOLT_TYPE_DISCRETE: discrete_levels is one word per level and holds
+ * num_levels words.
+ *
+ * RPMI_VOLT_TYPE_LINEAR: linear_levels is a flattened array of
+ * (uvolt_min, uvolt_max, uvolt_step) tuples, three words per level, and
+ * holds num_levels * 3 words.
  */
 struct rpmi_voltage_data {
 	/* voltage domain name */
@@ -2000,7 +2016,7 @@ struct rpmi_voltage_data {
 	rpmi_uint32_t				control;
 	/* regulator config */
 	rpmi_uint32_t				config;
-	/** number of supported voltage levels */
+	/** number of supported voltage levels (discrete levels or linear ranges) */
 	rpmi_uint32_t				num_levels;
 	/** transition latency */
 	rpmi_uint32_t				trans_latency;
@@ -2008,9 +2024,13 @@ struct rpmi_voltage_data {
         struct rpmi_voltage_discrete_range	*discrete_range;
         /** linear voltage range */
         struct rpmi_voltage_linear_range	*linear_range;
-	/** discrete voltage levels */
+	/** discrete voltage levels, one word per level, num_levels words */
 	rpmi_int32_t				*discrete_levels;
-	/** linear voltage levels */
+	/**
+	 * linear voltage levels, a flattened array of
+	 * (uvolt_min, uvolt_max, uvolt_step) tuples, three words per
+	 * level, num_levels * 3 words
+	 */
 	rpmi_int32_t				*linear_levels;
 	/** voltage level */
 	rpmi_int32_t				level_uv;
@@ -2028,7 +2048,12 @@ struct rpmi_voltage_attrs {
 	rpmi_uint32_t	num_levels;
 	/* transition latency */
 	rpmi_uint32_t	trans_latency;
-	/** array of supported levels */
+	/**
+	 * array of supported levels, either the discrete_levels or the
+	 * linear_levels array of struct rpmi_voltage_data selected by the
+	 * voltage format in capability, with the word-per-level layout
+	 * documented there
+	 */
 	rpmi_int32_t	*level_array;
 	/* voltage domain name */
 	const char	*name;
@@ -2065,7 +2090,14 @@ struct rpmi_voltage_platform_ops {
 				     rpmi_int32_t *volt_level);
 
 	/**
-	 * Get perf supported levels
+	 * Get voltage supported levels
+	 *
+	 * The max, volt_index and returned_levels parameters count voltage
+	 * levels, not words. The callback writes returned_levels levels into
+	 * level_array starting at level volt_index, which is
+	 * returned_levels words for a discrete domain and
+	 * returned_levels * 3 words for a linear domain, and never more
+	 * than max levels.
 	 **/
 	enum rpmi_error (*get_supp_levels)(void *priv,
 					   rpmi_uint32_t volt_id,
--- a/lib/rpmi_service_group_voltage.c
+++ b/lib/rpmi_service_group_voltage.c
@@ -9,22 +9,20 @@
 #include "librpmi_internal_list.h"
 
 #ifdef LIBRPMI_DEBUG
-#define DPRINTF(msg...)         rpmi_env_printf(msg)
+#define DPRINTF(msg...)		rpmi_env_printf(msg)
 #else
 #define DPRINTF(msg...)
 #endif
 
 #define RPMI_VOLT_NAME_MAX_LEN  16
 
-/** Convert list node pointer to struct rpmi_voltage instance pointer */
-#define to_rpmi_voltage(__node)    \
+/* Convert list node pointer to struct rpmi_voltage instance pointer */
+#define to_rpmi_voltage(__node)		\
 	container_of((__node), struct rpmi_voltage_data, node)
 
 /* A voltage domain instance */
 struct rpmi_voltage {
-	/* Lock to invoke the platform operations to
-	 * protect this structure
-	 */
+	/* Lock to invoke the platform operations to protect this structure */
 	void *lock;
 	/* domain ID */
 	rpmi_uint32_t id;
@@ -32,22 +30,22 @@ struct rpmi_voltage {
 	const struct rpmi_voltage_data *vdata;
 };
 
-/** RPMI Voltage Service Group instance */
+/* RPMI Voltage Service Group instance */
 struct rpmi_voltage_group {
 	/* Total domains count */
 	rpmi_uint32_t volt_count;
 	/* Pointer to voltage domains tree */
 	struct rpmi_voltage *volt_tree;
-	/* Common voltage platform operations (called with holding the lock)*/
+	/* Common voltage platform operations (called with holding the lock) */
 	const struct rpmi_voltage_platform_ops *ops;
 	/* Private data of platform voltage operations */
 	void *ops_priv;
 	struct rpmi_service_group group;
 };
 
-/** Get a struct rpmi_voltage instance pointer from voltage domain id */
-static inline struct rpmi_voltage *
-rpmi_get_voltage(struct rpmi_voltage_group *volt_group, rpmi_uint32_t volt_id)
+/* Get a struct rpmi_voltage instance pointer from voltage domain id */
+static inline struct rpmi_voltage
+*rpmi_get_voltage(struct rpmi_voltage_group *volt_group, rpmi_uint32_t volt_id)
 {
 	return &volt_group->volt_tree[volt_id];
 }
@@ -127,7 +125,6 @@ static enum rpmi_error __rpmi_volt_get_config(struct rpmi_voltage_group *voltgrp
 	rpmi_env_lock(volt->lock);
 
 	ret = voltgrp->ops->get_config(voltgrp->ops_priv, volt->id, &config);
-
 	*volt_config = config;
 
 	rpmi_env_unlock(volt->lock);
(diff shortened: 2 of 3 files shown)


lib/rpmi_service_group_voltage.c
```diff
@@ -299,6 +299,8 @@ rpmi_volt_get_attributes(struct rpmi_service_group *group,
 	if (volt_attrs.name)
 		rpmi_env_strncpy((char *)&resp[4], volt_attrs.name, RPMI_VOLT_NAME_MAX_LEN);
 
+	resp[0] = rpmi_to_xe32(trans->is_be, (rpmi_uint32_t)RPMI_SUCCESS);
+
 	resp_dlen = 4 * sizeof(*resp) + RPMI_VOLT_NAME_MAX_LEN;
 
 done:
```
saini-ranbirs, line 293: `resp[0] = rpmi_to_xe32(trans->is_be, (rpmi_uint32_t)ret);`

This common statement can be added here which can serve all paths with `ret` filled with appropriate value earlier.

lib/rpmi_service_group_voltage.c
saini-ranbirs: Not directly related to the change set, but the response statements do not seem to need blank lines in between which for me do not add any visual value too.

saini-ranbirs (commented):
Commit message
      hw/misc/librpmi => No such tree structure

voltage: ... should be good enough

lib/rpmi_service_group_voltage.c
```diff
@@ -456,7 +468,7 @@ rpmi_volt_get_supp_levels(struct rpmi_service_group *group,
                 goto done;
         }
 
-        for (i = 0; i < returned; i++) {
+        for (i = 0; i < returned * words_per_level; i++) {
                 start = 4;
```
saini-ranbirs: same assignment inside loop?

`4 + i` can be used simply

lib/rpmi_service_group_voltage.c
```diff
@@ -471,7 +483,7 @@ rpmi_volt_get_supp_levels(struct rpmi_service_group *group,
         resp[0] = rpmi_to_xe32(trans->is_be, (rpmi_uint32_t)RPMI_SUCCESS);
 
         resp_dlen = (4 * sizeof(*resp)) +
-                        (returned * sizeof(rpmi_uint32_t));
+                        (returned * words_per_level * sizeof(rpmi_uint32_t));
 
 done:
         *response_datalen = resp_dlen;
```
saini-ranbirs: `resp[0] = rpmi_to_xe32(trans->is_be, (rpmi_uint32_t)ret);`

This common statement can be added here which can serve all paths with `ret` filled with appropriate value earlier.

include/librpmi.h
```diff
@@ -1945,6 +1945,10 @@ enum rpmi_voltage_type {
 	(((__format) & RPMI_VOLT_CAPABILITY_FORMAT_MASK) <<		\
 	 RPMI_VOLT_CAPABILITY_FORMAT_POS)
 
+#define RPMI_VOLT_CAPABILITY_GET_FORMAT(__capability)			\
+	(((__capability) >> RPMI_VOLT_CAPABILITY_FORMAT_POS) &		\
```
slingappa, line 1962: This adds the format getter and the service now sizes/copies linear replies as 3 words per level, but the public `struct rpmi_voltage_data` comments below still do not document the layout contract. Please spell out that `num_levels`, `max`, and `returned` count voltage levels/ranges, while `linear_levels` is a flattened `(min, max, step)` tuple array and `discrete_levels` remains one word per level. Without that, platform users can keep providing the old one-word-per-level linear array and get malformed replies or undersized buffers.

lib/rpmi_service_group_voltage.c
```diff
@@ -387,94 +375,101 @@ rpmi_volt_set_config(struct rpmi_service_group *group,
...
+		ret = RPMI_ERR_NOTSUPP;
+		goto done;
+	}
+
+	volt_level_idx = rpmi_to_xe32(trans->is_be,
+				      ((const rpmi_uint32_t *)request_data)[1]);
+
+	/*
+         * A discrete level is one word while a linear range level is a
+         * (min, max, step) tuple of three. NUM_LEVELS, RETURNED and REMAINING
```
slingappa: Checkpatch still reports this added block comment as `ERROR: code indent should use tabs where possible` for lines 432-436. Please align the `*` lines with the surrounding tab-based indentation before merging.

lib/rpmi_service_group_voltage.c
```diff
@@ -387,94 +375,101 @@ rpmi_volt_set_config(struct rpmi_service_group *group,
 	resp_dlen = sizeof(*resp);
 
 done:
+	resp[0] = rpmi_to_xe32(trans->is_be, (rpmi_uint32_t)ret);
 	*response_datalen = resp_dlen;
```
slingappa: This unconditional `done:` write makes the earlier success-path `resp[0] = ... RPMI_SUCCESS` assignment redundant. Please drop that earlier assignment so `rpmi_volt_set_config()` follows the same single-status-write pattern as the other updated voltage handlers.

test/test_srvgrp_voltage.c
```diff
@@ -154,13 +244,82 @@ static rpmi_uint32_t get_mem_levels_reqdata[] = {
 	0,
 };
 
+/* One tuple returned, none remaining, carrying all three of its words. */
+/* One tuple returned, none remaining, carrying all three of its words. */
```
slingappa, line 247: This explanatory comment is duplicated back-to-back. Please remove one copy.

slingappa (commented):
left few more comments.

lib/rpmi_service_group_voltage.c
```diff
@@ -362,16 +359,14 @@ rpmi_volt_set_config(struct rpmi_service_group *group,
 					((const rpmi_uint32_t *)request_data)[1]);
 
 	if (voltid >= voltgrp->volt_count) {
-		resp[0] = rpmi_to_xe32(trans->is_be,
-				       (rpmi_uint32_t)RPMI_ERR_INVALID_PARAM);
+		ret = RPMI_ERR_INVALID_PARAM;
 		resp_dlen = sizeof(*resp);
```
saini-ranbirs: See if even the `resp_dlen` or  `*response_datalen` can be set commonly after the `done:` label

If the success and failure paths have same response data length, then just set -
        `*response_datalen = common value;`

If the success and failure paths have different response data length, then set length appropriately as in -
        `*response_datalen = (ret) ? failure path value; success path value;`
or if (ret) ... else ... approach.

I doubt there will be cases of different response data length values for different failure paths.

lib/rpmi_service_group_voltage.c
saini-ranbirs: Now that this commit includes some general cosmetic issues, let me point few more issues that can be addressed in this context

lock)*/ => lock) */

saini-ranbirs: https://docs.kernel.org/process/coding-style.html

When declaring pointer data or a function that returns a pointer type, the preferred use of * is adjacent to the data name or function name and not adjacent to the type name.

saini-ranbirs: Same comment as at Line 49

saini-ranbirs: misaligned

saini-ranbirs: misaligned

lib/rpmi_service_group_voltage.c
```diff
@@ -388,93 +381,93 @@ rpmi_volt_set_config(struct rpmi_service_group *group,
...
-                resp[start + i] = rpmi_to_xe32(trans->is_be,
-                                        (rpmi_uint32_t)volt_level_array[i]);
-        }
+	rpmi_int32_t *volt_level_array;
+	struct rpmi_voltage_attrs volt_attrs;
+	struct rpmi_voltage_group *voltgrp = group->priv;
+	rpmi_uint32_t *resp = (void *)response_data;
+
+	rpmi_uint32_t voltid = rpmi_to_xe32(trans->is_be,
+				((const rpmi_uint32_t *)request_data)[0]);
```
saini-ranbirs: misaligned

lib/rpmi_service_group_voltage.c
```diff
@@ -428,10 +428,20 @@ rpmi_volt_get_supp_levels(struct rpmi_service_group *group,
...
 
+	/*
+         * A discrete level is one word while a linear range level is a
+         * (min, max, step) tuple of three. NUM_LEVELS, RETURNED and REMAINING
+         * all count levels, so the width of a level is needed both to size the
+         * reply and to know how many levels fit in one message.
+         */
+	words_per_level =
+		RPMI_VOLT_CAPABILITY_GET_FORMAT(volt_attrs.capability) ==
+			RPMI_VOLT_TYPE_LINEAR ? 3 : 1;
```
saini-ranbirs: move at same level as last line

lib/rpmi_service_group_voltage.c
```diff
@@ -428,10 +428,20 @@ rpmi_volt_get_supp_levels(struct rpmi_service_group *group,
...
+
 	/* max volt levels a rpmi message can accommodate */
 	max_levels =
 		(RPMI_MSG_DATA_SIZE(trans->slot_size) - (4 * sizeof(*resp))) /
-					sizeof(rpmi_int32_t);
+				    (sizeof(rpmi_int32_t) * words_per_level);
```
saini-ranbirs: move at same level as last line

lib/rpmi_service_group_voltage.c
```diff
@@ -440,9 +450,8 @@ rpmi_volt_get_supp_levels(struct rpmi_service_group *group,
 		goto done;
 	}
 
-	for (i = 0; i < returned; i++) {
-		start = 4;
-		resp[start + i] = rpmi_to_xe32(trans->is_be,
+	for (i = 0; i < returned * words_per_level; i++) {
+		resp[4 + i] = rpmi_to_xe32(trans->is_be,
 					(rpmi_uint32_t)volt_level_array[i]);
```
saini-ranbirs: misaligned now

lib/rpmi_service_group_voltage.c
```diff
@@ -388,93 +381,93 @@ rpmi_volt_set_config(struct rpmi_service_group *group,
...
+		goto done;
+	}
+
+	for (i = 0; i < returned; i++) {
+		start = 4;
+		resp[start + i] = rpmi_to_xe32(trans->is_be,
+					(rpmi_uint32_t)volt_level_array[i]);
+	}
 
        remaining = num_volt_level - (volt_level_idx + returned);
```
saini-ranbirs: tab to be used at the beginning and not spaces

lib/rpmi_service_group_voltage.c
saini-ranbirs: misaligned

saini-ranbirs: misaligned

saini-ranbirs: Same comment as at Line 49

saini-ranbirs: can fit in previous line within limits

saini-ranbirs: can fit in previous line within limits

include/librpmi.h
yeongjoshua: ok, ill add comments in `struct rpmi_voltage_data`

yeongjoshua:
Thanks @saini-ranbirs for review comments

lib/rpmi_service_group_voltage.c
saini-ranbirs: Seems this has been missed. If intentional, then just let us know.

saini-ranbirs: I realized that common handling wrt response data length has been taken care of in `rpmi_volt_set_config()`.

lib/rpmi_service_group_voltage.c
```diff
@@ -360,25 +356,15 @@ rpmi_volt_set_config(struct rpmi_service_group *group,
...
-	}
-
-	resp[0] = rpmi_to_xe32(trans->is_be, (rpmi_uint32_t)RPMI_SUCCESS);
-
-	resp_dlen = sizeof(*resp);
 
 done:
-	*response_datalen = resp_dlen;
+	resp[0] = rpmi_to_xe32(trans->is_be, (rpmi_uint32_t)ret);
+	*response_datalen = sizeof(*resp);
```
saini-ranbirs, line 367: Just like common handling wrt response data length has been done `rpmi_volt_set_config()`, see if it is good to extend the same approach elsewhere in the file.

If the success and failure paths have different response data length, then set length appropriately as in -
`*response_datalen = (ret) ? failure path value; success path value;`
or `if (ret) ... else ...` approach.

yeongjoshua: thanks, i updated this part

saini-ranbirs (approved):

PR #119, opened 2026-08-31
cppc: advertise optional fast-channel doorbell (follow-up to autonomous mode)

icosta-sifive:
This is a follow-up to the previous [PR](https://github.com/riscv-software-src/librpmi/pull/115) that added CPPC autonomous mode. While wiring up `GET_FAST_CHANNEL_REGION` for autonomous mode, one commit was forgotten. The optional fast-channel doorbell that the same service advertise, this PR fills that gap.

Until now librpmi hardcoded these fields to zero, so even with the rest of the fast-channel path in place a platform had no way to advertise a doorbell.

The doorbell lets the application processor signal the platform microcontroller
after updating its Perf Request fast-channel, instead of relying on polling. It
is described by:

librpmi was still hardcoding these fields to zero, so a platform could never advertise a doorbell even though the rest of the fast-channel path was in place.

## Tests

Added test coverage for:
- Doorbell advertisement in passive mode (32-bit doorbell)
- Doorbell advertisement in autonomous mode (combined `FLAGS[4:3]` + doorbell bits)
- S-mode access to `GET_FAST_CHANNEL_REGION`

### DIFF
--- a/include/librpmi.h
+++ b/include/librpmi.h
@@ -1438,6 +1438,26 @@ enum rpmi_cppc_mode {
 /**
  * ACPI CPPC Registers
  */
+/** CPPC fastchannel flags */
+#define RPMI_CPPC_FST_CHN_DB_REG_08_BITS	(0U << 1)
+#define RPMI_CPPC_FST_CHN_DB_REG_16_BITS	(1U << 1)
+#define RPMI_CPPC_FST_CHN_DB_REG_32_BITS	(2U << 1)
+
+#define RPMI_CPPC_FST_CHN_DB_NOT_SUPP		(0U << 0)
+#define RPMI_CPPC_FST_CHN_DB_SUPP		(1U << 0)
+
+/** CPPC Perf Request fast-channel doorbell */
+struct rpmi_cppc_fastchan_doorbell {
+	/** doorbell flags */
+	rpmi_uint32_t flags;
+	/** doorbell addr low */
+	rpmi_uint32_t db_addr_low;
+	/** doorbell addr high */
+	rpmi_uint32_t db_addr_high;
+	/** doorbell write value */
+	rpmi_uint32_t db_write_value;
+};
+
 struct rpmi_cppc_regs {
 	/* highest performance (r) */
 	rpmi_uint32_t highest_perf;
@@ -1483,6 +1503,8 @@ struct rpmi_cppc_regs {
 	rpmi_uint32_t nominal_freq;
 	/* transition latency */
 	rpmi_uint32_t transition_latency;
+	/* Perf Request fast-channel doorbell (optional) */
+	struct rpmi_cppc_fastchan_doorbell doorbell;
 };
 
 /* CPPC Fastchannel size of both types as per RPMI spec */
--- a/lib/rpmi_service_group_cppc.c
+++ b/lib/rpmi_service_group_cppc.c
@@ -637,36 +637,50 @@ rpmi_cppc_sg_get_fast_channel_region(struct rpmi_service_group *group,
 	rpmi_uint32_t resp_dlen, flags;
 	rpmi_uint64_t fastchan_region_base, fastchan_region_size;
 	struct rpmi_cppc_group *cppcgrp = group->priv;
+	struct rpmi_cppc_fastchan *fc = cppcgrp->fastchan_ctx;
+	const struct rpmi_cppc_fastchan_doorbell *db = &cppcgrp->regs->doorbell;
 	rpmi_uint32_t *resp = (void *)response_data;
 
-	if (!cppcgrp->fastchan_ctx) {
+	if (!fc) {
 		status = RPMI_ERR_NOTSUPP;
 		resp_dlen = sizeof(*resp);
 		goto done;
 	}
 
-	fastchan_region_base = rpmi_shmem_base(cppcgrp->fastchan_ctx->shmem);
-	fastchan_region_size = rpmi_shmem_size(cppcgrp->fastchan_ctx->shmem);
+	fastchan_region_base = rpmi_shmem_base(fc->shmem);
+	fastchan_region_size = rpmi_shmem_size(fc->shmem);
 
-	/* FLAGS[4:3]: 0b00 = normal/passive, 0b01 = autonomous. No doorbell. */
+	/* FLAGS[4:3]: 0b00 = normal/passive, 0b01 = autonomous. */
 	flags = (cppcgrp->cppc_mode == RPMI_CPPC_AUTO_MODE) ? (0x01U << 3) : 0;
+
+	/* FLAGS[2:1] doorbell width, FLAGS[0] doorbell support (optional) */
+	if (db->flags & RPMI_CPPC_FST_CHN_DB_SUPP) {
+		flags |= RPMI_CPPC_FST_CHN_DB_SUPP | (db->flags & (0x3U << 1));
+		/* doorbell addr low */
+		resp[6] = rpmi_to_xe32(trans->is_be, db->db_addr_low);
+		/* doorbell addr high */
+		resp[7] = rpmi_to_xe32(trans->is_be, db->db_addr_high);
+		/* doorbell write value */
+		resp[8] = rpmi_to_xe32(trans->is_be, db->db_write_value);
+	} else {
+		/* doorbell addr low */
+		resp[6] = 0;
+		/* doorbell addr high */
+		resp[7] = 0;
+		/* doorbell write value */
+		resp[8] = 0;
+	}
 
 	status = RPMI_SUCCESS;
 	resp[1] = rpmi_to_xe32(trans->is_be, (rpmi_uint32_t)flags);
 	/* fast channel region address low */
 	resp[2] = rpmi_to_xe32(trans->is_be, (rpmi_uint32_t)fastchan_region_base);
-	/* fast channel region address low */
+	/* fast channel region address high */
 	resp[3] = rpmi_to_xe32(trans->is_be, (rpmi_uint32_t)(fastchan_region_base >> 32));
 	/* fast channel region size low */
-	resp[4] = rpmi_to_xe32(trans->is_be, (rpmi_uint32_t)fastchan_region_size);;
+	resp[4] = rpmi_to_xe32(trans->is_be, (rpmi_uint32_t)fastchan_region_size);
 	/* fast channel region size high */
 	resp[5] = rpmi_to_xe32(trans->is_be, (rpmi_uint32_t)(fastchan_region_size >> 32));
-	/* doorbell addr low */
-	resp[6] = 0;
-	/* doorbell addr high */
-	resp[7] = 0;
-	/* doorbell write value */
-	resp[8] = 0;
 
 	resp_dlen = 9 * sizeof(*resp);
 
--- a/test/test_srvgrp_cppc.c
+++ b/test/test_srvgrp_cppc.c
@@ -29,6 +29,14 @@
 #define TEST_CPPC_AUTO_MIN_PERF			0x20
 #define TEST_CPPC_AUTO_MAX_PERF			0x40
 
+/* Doorbell scenario: passive mode + 32-bit doorbell */
+#define TEST_CPPC_DB_ADDR_LOW			0x30000000
+#define TEST_CPPC_DB_ADDR_HIGH			0x0
+#define TEST_CPPC_DB_WRITE_VALUE		0x7
+/* FLAGS[2:1]=0b10 (32-bit) | FLAGS[0]=1 (doorbell supported) */
+#define TEST_CPPC_DB_FLAGS			(RPMI_CPPC_FST_CHN_DB_REG_32_BITS | \
+						 RPMI_CPPC_FST_CHN_DB_SUPP)
+
 static rpmi_uint32_t test_auto_captured_min_perf;
 static rpmi_uint32_t test_auto_captured_max_perf;
 
@@ -880,7 +888,7 @@ static struct rpmi_test_scenario scenario_cppc_s_mode = {
 	.init = test_cppc_auto_scenario_init,
 	.cleanup = test_scenario_default_cleanup,
 
-	.num_tests = 4,
+	.num_tests = 5,
 	.tests = {
 		{
 			.name = "PROBE HIGHEST_PERF (S-mode access)",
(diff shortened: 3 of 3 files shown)


icosta-sifive:
@vlsunil FYI

lib/rpmi_service_group_cppc.c
```diff
@@ -637,36 +637,50 @@ rpmi_cppc_sg_get_fast_channel_region(struct rpmi_service_group *group,
...
+		/* doorbell write value */
+		resp[8] = rpmi_to_xe32(trans->is_be, db->db_write_value);
+	} else {
+		/* doorbell addr low */
+		resp[6] = 0;
+		/* doorbell addr high */
+		resp[7] = 0;
+		/* doorbell write value */
+		resp[8] = 0;
+	}
```
saini-ranbirs, line 670: The comments for `resp[6/7/8]` are visible in close proximity through the if part, so else part even without comments wrt `resp[6/7/8]` and the `resp[6] = resp[7]= resp[8] = 0;` assignment should be good enough.

lib/rpmi_service_group_cppc.c
```diff
@@ -637,36 +637,45 @@ rpmi_cppc_sg_get_fast_channel_region(struct rpmi_service_group *group,
...
+	fastchan_region_base = rpmi_shmem_base(fc->shmem);
+	fastchan_region_size = rpmi_shmem_size(fc->shmem);
 
-	/* FLAGS[4:3]: 0b00 = normal/passive, 0b01 = autonomous. No doorbell. */
+	/* FLAGS[4:3]: 0b00 = normal/passive, 0b01 = autonomous. */
 	flags = (cppcgrp->cppc_mode == RPMI_CPPC_AUTO_MODE) ? (0x01U << 3) : 0;
 
+	/* FLAGS[2:1] doorbell width, FLAGS[0] doorbell support (optional) */
+	if (db->flags & RPMI_CPPC_FST_CHN_DB_SUPP) {
+		flags |= RPMI_CPPC_FST_CHN_DB_SUPP | (db->flags & (0x3U << 1));
```
saini-ranbirs: See if it's better to define `(0x3U << 1)` as a macro - say as `RPMI_CPPC_FST_CHN_DB_REG_WIDTH_MASK`?

and may I know why `flags |= db->flags;` is not sufficient here, I mean what else `db->flags` contain?

lib/rpmi_service_group_cppc.c
```diff
@@ -637,36 +637,45 @@ rpmi_cppc_sg_get_fast_channel_region(struct rpmi_service_group *group,
...
 	}
 
-	fastchan_region_base = rpmi_shmem_base(cppcgrp->fastchan_ctx->shmem);
-	fastchan_region_size = rpmi_shmem_size(cppcgrp->fastchan_ctx->shmem);
+	fastchan_region_base = rpmi_shmem_base(fc->shmem);
+	fastchan_region_size = rpmi_shmem_size(fc->shmem);
 
-	/* FLAGS[4:3]: 0b00 = normal/passive, 0b01 = autonomous. No doorbell. */
+	/* FLAGS[4:3]: 0b00 = normal/passive, 0b01 = autonomous. */
 	flags = (cppcgrp->cppc_mode == RPMI_CPPC_AUTO_MODE) ? (0x01U << 3) : 0;
```
saini-ranbirs: This of my review comment is out of context, but still can you analyze what does it take to just set `flags = cppcgrp->cppc_mode;` here?
 
Thought: `RPMI_CPPC_AUTO_MODE` can be defined bitwise and not enum like

`#define RPMI_CPPC_AUTO_MODE  (0x01U << 3)`

If it seems worth, then this can be taken up separately.

lib/rpmi_service_group_cppc.c
icosta-sifive: Added `RPMI_CPPC_FST_CHN_DB_REG_WIDTH_MASK`. Although flags |= db->flags works when the platform provides a valid value, `rpmi_cppc_fastchan_doorbell.flags` is an unrestricted 32-bit field. Explicitly copying only bits [2:0] ensures invalid platform data cannot corrupt the independently generated mode bits [4:3] or set reserved bits [31:5], which the RPMI specification requires to be zero. I think keeping the explicit mask is safer, but please let me know what you think. I can switch to `flags |= db->flags` if you prefer the simpler approach.

lib/rpmi_service_group_cppc.c
icosta-sifive: I analyzed this and the current implementation is spec-compliant. I would prefer to keep enum rpmi_cppc_mode as the semantic API value and encode it into FLAGS only when building the response. Defining `RPMI_CPPC_AUTO_MODE as (0x01U << 3)` would couple the public mode value to this specific wire layout and change its value from 1 to 8. We could introduce a named macro for the FLAGS encoding as a separate cleanup, if useful.

lib/rpmi_service_group_cppc.c
saini-ranbirs: So, does that mean there is no validation of `db->flags` when the platform provides an invalid value?

icosta-sifive: Correct, there is currently no explicit validation of db->flags; the mask only prevents invalid bits from reaching the wire. On reconsideration, validating the platform data during `rpmi_service_group_cppc_create()` would be cleaner. Once validated, flags |= db->flags would be sufficient. I can update the implementation accordingly if you agree ?

saini-ranbirs: To me, validating the platform data during `rpmi_service_group_cppc_create()` is better. It should allow group creation with valid set of values only.

lib/rpmi_service_group_cppc.c
saini-ranbirs: I think macros should be used wrt FLAGS[4:3] instead of using `(0x01U << 3) : 0` so that code can be self-descriptive.

lib/rpmi_service_group_cppc.c
icosta-sifive: Agreed, that makes sense. I’ve updated rpmi_service_group_cppc_create() to reject invalid doorbell flags during group creation. With that validation in place, the response path now simply uses flags |= db->flags. I also added tests for the invalid cases.

include/librpmi.h
```diff
@@ -1438,6 +1438,30 @@ enum rpmi_cppc_mode {
 /**
  * ACPI CPPC Registers
  */
+/** CPPC fastchannel flags */
+#define RPMI_CPPC_FST_CHN_MODE_NORMAL		(0U << 3)
+#define RPMI_CPPC_FST_CHN_MODE_AUTONOMOUS	(1U << 3)
+
+#define RPMI_CPPC_FST_CHN_DB_REG_WIDTH_MASK	(0x3U << 1)
```
vlsunil: Can you be consistent to add 0x for all macros?

include/librpmi.h
```diff
@@ -1438,6 +1438,30 @@ enum rpmi_cppc_mode {
...
+#define RPMI_CPPC_FST_CHN_DB_REG_08_BITS	(0U << 1)
```
vlsunil: 1. Can you be consistent to add 0x for all macros?
2. Isn't it better to define BIT and GENMASK?

include/librpmi.h
```diff
@@ -1438,6 +1438,30 @@ enum rpmi_cppc_mode {
...
+#define RPMI_CPPC_FST_CHN_DB_REG_16_BITS	(1U << 1)
+#define RPMI_CPPC_FST_CHN_DB_REG_32_BITS	(2U << 1)
+
+#define RPMI_CPPC_FST_CHN_DB_NOT_SUPP		(0U << 0)
+#define RPMI_CPPC_FST_CHN_DB_SUPP		(1U << 0)
+
+/** CPPC Perf Request fast-channel doorbell */
+struct rpmi_cppc_fastchan_doorbell {
+	/** doorbell flags */
```
vlsunil: It is not clear to me why Flags is made part of doorbell structure. While Flags has doorbell related info, it has other info as well, right?

test/test_srvgrp_cppc.c
```diff
@@ -938,6 +946,267 @@ static struct rpmi_test_scenario scenario_cppc_s_mode = {
...
+		.db_write_value = TEST_CPPC_DB_WRITE_VALUE,
+	},
+};
+
+static rpmi_uint16_t init_db_fastchan_region_expected(struct rpmi_test_scenario *scene,
+						      struct rpmi_test *test,
+						      void *data,
+						      rpmi_uint16_t max_data_len)
+{
+	rpmi_uint32_t *exp = data;
```
vlsunil, line 1009: I believe librpmi follows linux coding standards/formatting. Please check.

include/librpmi.h
```diff
@@ -1483,6 +1507,8 @@ struct rpmi_cppc_regs {
 	rpmi_uint32_t nominal_freq;
 	/* transition latency */
 	rpmi_uint32_t transition_latency;
+	/* Perf Request fast-channel doorbell (optional) */
+	struct rpmi_cppc_fastchan_doorbell doorbell;
```
vlsunil, line 1511: Again, can you explain why it is added here?

include/librpmi.h
icosta-sifive: Updated the new macros to use hexadecimal values consistently. I kept the explicit shifts rather than BIT()/GENMASK() because those helpers are not currently provided by librpmi and this is a public header. Introducing generic helper macros here could conflict with the integrating firmware, while the explicit expressions match the existing header style.

include/librpmi.h
icosta-sifive: This flags member contains only the doorbell-related fields: support in bit 0 and register width in bits [2:1]. It does not contain the complete response FLAGS value; CPPC mode in bits [4:3] is obtained separately from cppc_mode and combined when building the response. I’ve updated the member comment to clarify this.

include/librpmi.h
icosta-sifive: I added it here because cppc_regs is the existing container for static CPPC platform data retained by the service group. The doorbell configuration has the same lifetime, and embedding it here avoids changing the rpmi_service_group_cppc_create() signature. I agree that the rpmi_cppc_regs name makes this less obvious. If you prefer stricter separation, I can pass the doorbell descriptor separately, but that would require changing the public create API.

test/test_srvgrp_cppc.c
icosta-sifive: Good point. I reformatted the new test functions and ran the complete patch through Linux checkpatch.pl --strict; it now passes with 0 errors, 0 warnings, and 0 checks.

lib/rpmi_service_group_cppc.c
icosta-sifive: @saini-ranbirs I initially combined these assignments for conciseness, but reverted to separate assignments after running the full patch through Linux checkpatch.pl --strict, which flags chained assignments with MULTIPLE_ASSIGNMENTS. The behavior is unchanged; this is only to follow the project’s Linux-style formatting and keep the patch checkpatch-clean.

saini-ranbirs: That's fine. However, the else part is for DB not supported. So, see if those comments there really make sense.

icosta-sifive: Agreed. Since the else path means the doorbell is not supported, the individual doorbell field comments are unnecessary there. I’ve removed them and kept the assignments separate to avoid the checkpatch MULTIPLE_ASSIGNMENTS warning.

include/librpmi.h
vlsunil: I personally don't prefer to add fast channel related information here. Can it be added under rpmi_cppc_fastchan structure since this is fast channel specific information? Would that change the API?

lib/rpmi_service_group_cppc.c
```diff
@@ -1027,6 +1039,19 @@ rpmi_service_group_cppc_create(struct rpmi_hsm *hsm,
 		return NULL;
 	}
 
+	if (cppc_regs->doorbell.flags &
```
slingappa, line 1042: The creation-time validation rejects reserved bits and width encoding 3, but it still accepts a nonzero width with RPMI_CPPC_FST_CHN_DB_SUPP clear (for example, RPMI_CPPC_FST_CHN_DB_REG_32_BITS). The current response path drops those width bits, so this is not a wire-level defect today, but it leaves the public descriptor semantically ambiguous. Could we reject nonzero width when support is clear, or add a test/documentation showing that the specification explicitly permits those bits to be ignored?

test/test_srvgrp_cppc.c
```diff
@@ -938,6 +946,267 @@ static struct rpmi_test_scenario scenario_cppc_s_mode = {
...
+static struct rpmi_cppc_regs test_cppc_db_regs = {
+	.highest_perf = TEST_CPPC_HIGHEST_PERF,
+	.nominal_perf = TEST_CPPC_NOMINAL_PERF,
+	.lowest_nonlinear_perf = 0x30,
+	.lowest_perf = 0x10,
+	.reference_perf = TEST_CPPC_NOMINAL_PERF,
+	.lowest_freq = 1000000000U,
+	.nominal_freq = 2000000000U,
+	.transition_latency = 10,
+	.doorbell = {
```
slingappa, line 986: Both new positive scenarios use only RPMI_CPPC_FST_CHN_DB_REG_32_BITS, but the public API defines three legal doorbell widths (8, 16, and 32 bits), and the implementation validates the width field. Could we add cases for the 8-bit and 16-bit encodings as well—ideally table-driven—while keeping the reserved-width rejection test? Otherwise a regression in FLAGS[2:1] for either legal variant would leave the suite green.

lib/rpmi_service_group_cppc.c
saini-ranbirs: As per spec, the doorbell details are unspecified and considered invalid if the Performance Request fast-channel
doorbell (FLAGS[0] = 0) is not supported and must not be used.

Hence, it seems safe even if else part assignments are totally removed. One more option is to set the `resp_dlen` appropriately based on DB supported or not supported as 9 or 6 times .. respectively.

icosta-sifive: Although those fields must not be used when the doorbell is unsupported, librpmi reuses the ACK buffer, so leaving words 6–8 unwritten could expose stale data. Also, the spec defines words 0–8 as the response layout, so I’d prefer to keep the fixed 9-word response and explicitly zero the unused fields.

lib/rpmi_service_group_cppc.c
icosta-sifive: Good catch. I’ve updated the creation-time validation to reject a nonzero doorbell width when RPMI_CPPC_FST_CHN_DB_SUPP is clear, and added a negative test for this case. This keeps the public descriptor unambiguous: when the doorbell is unsupported, its flags must be zero.

test/test_srvgrp_cppc.c
icosta-sifive: Good point. I’ve added table-driven positive tests for the 8-, 16-, and 32-bit doorbell width encodings, while retaining the reserved-width rejection test. This now covers every legal value of FLAGS[2:1].

include/librpmi.h
icosta-sifive: Yes, that would require an API change because `struct rpmi_cppc_fastchan` is private and created internally, so the platform currently has no way to populate the doorbell information there. We would need to pass a doorbell descriptor, or a new public fast-channel configuration structure, to `rpmi_service_group_cppc_create()`, then store it in the private fast-channel context. I agree that this would provide cleaner ownership. If changing the public API is acceptable, I can rework it that way.

lib/rpmi_service_group_cppc.c
```diff
@@ -1027,6 +1039,26 @@ rpmi_service_group_cppc_create(struct rpmi_hsm *hsm,
...
+	}
+
+	if ((cppc_regs->doorbell.flags & RPMI_CPPC_FST_CHN_DB_REG_WIDTH_MASK) ==
+	    RPMI_CPPC_FST_CHN_DB_REG_WIDTH_MASK) {
+		DPRINTF("%s: invalid cppc fastchannel doorbell width\n", __func__);
+		return NULL;
+	}
+
+	if (!(cppc_regs->doorbell.flags & RPMI_CPPC_FST_CHN_DB_SUPP) &&
+	    (cppc_regs->doorbell.flags & RPMI_CPPC_FST_CHN_DB_REG_WIDTH_MASK)) {
```
saini-ranbirs: Shouldn't this validation be done before the cppc fastchannel doorbell width validation?

include/librpmi.h
slingappa: I agree with @vlsunil here, I prefer passing an optional `const struct rpmi_cppc_fastchan_doorbell *` to `rpmi_service_group_cppc_create()`

icosta-sifive: @slingappa @vlsunil In that case i will proceed with this changes even knowing that brakes API compatibility here

lib/rpmi_service_group_cppc.c
icosta-sifive: Agreed. Checking for a width without doorbell support first makes the validation flow clearer, so I’ve reordered the checks accordingly.

include/librpmi.h
icosta-sifive: @slingappa @vlsunil As agreed, i’ve changed the API to pass an optional const struct rpmi_cppc_fastchan_doorbell * to rpmi_service_group_cppc_create(). The descriptor is now stored in the private fast-channel context instead of rpmi_cppc_regs; NULL represents no doorbell. All call sites and tests have been updated.

